PR #1, opened 2017-02-14
strum_macros: absolutize type paths

durka:
This commit changes all the type paths in quoted code to be absolute, so the expanded code works in modules and in case stdlib types are shadowed (e.g. a local `Result` alias).

### DIFF
--- a/strum_macros/src/lib.rs
+++ b/strum_macros/src/lib.rs
@@ -111,7 +111,7 @@ fn from_string_inner(ast: &syn::DeriveInput) -> quote::Tokens {
     };
 
     let mut has_default = false;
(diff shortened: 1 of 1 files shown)


Peternator7:
These changes look good! Thanks for the PR

durka:
Awesome! When you have a chance could you publish a new version so I can put my crate (that relies on this fix) up on crates.io?

Peternator7:
I'll get it pushed tonight, and add a comment here when it's published!

Peternator7:
strum v0.2.0 is out.  Thanks for your contribution :)

durka:
foreman v0.2.1 is out!

PR #8, opened 2017-06-30
Add some missing absolute module paths

kyren:
I was getting:

```Use of undeclared type or module `strum` ``` and

```Use of undeclared type or module `std` ```

so I dug into the source and found these.

I'm not super familiar with procedural macros, I just made them follow the pattern for the rest of the file, apologies if there's subtlety I'm missing.


Peternator7:
This looks good! I'll try and get an update pushed to https://crates.io this evening. Thanks for the fix :)

Issue #9, opened 2017-08-19
Implement `AsRef<str>`?


lo48576:
`AsRef<str>` provides string without memory allocation, so I want it.

lo48576:
Implemented by #10.

Peternator7:
Updated version is on [crates.io](https://crates.io/crates/strum_macros)

PR #10, opened 2017-08-19
Add `derive(AsRefStr)` implementation

lo48576:
* `derive(AsRefStr)` implements `AsRef<str>` (UPDATED: 558cdd0cd018dfe3a1b4fd8f8636ab514ac4cfd7)
* ~~`derive(AsStr)` uses `as_str`, `to_string`, `serialize` attributes (in this order).~~
* ~~`derive(ToString)` uses `to_string`, `as_str`, `serialize` attributes (in this order).~~
* `derive(AsRefStr)` uses `to_string` and `serialize` attributes (in this order, same as `derive(ToString)`). (UPDATED: a1b480dfc8c5e0bae9e07aed8821da4f1710383c, 558cdd0cd018dfe3a1b4fd8f8636ab514ac4cfd7)

I don't know which is better: `AsRef<str> for Enum` vs. `Enum::as_str()`, but I used `AsRef<str>` because it is similar to `ToString` (both are traits and therefore non-invasive).

Also I can't decide which name is better, `AsStr` vs. `AsRefStr`.

README documentation is not updated in this PR...

This solves #9.

### DIFF
--- a/strum_macros/src/as_str.rs
+++ b/strum_macros/src/as_str.rs
@@ -1,0 +1,62 @@
+
+use quote;
+use syn;
+
+use helpers::{unique_attr, extract_attrs, is_disabled};
+
+pub fn as_str_inner(ast: &syn::DeriveInput) -> quote::Tokens {
+    let name = &ast.ident;
+    let (impl_generics, ty_generics, where_clause) = ast.generics.split_for_impl();
+    let variants = match ast.body {
+        syn::Body::Enum(ref v) => v,
+        _ => panic!("AsStr only works on Enums"),
+    };
+
+    let mut arms = Vec::new();
+    for variant in variants {
+        use syn::VariantData::*;
+        let ident = &variant.ident;
+
+        if is_disabled(&variant.attrs) {
+            continue;
+        }
+
+        // Look at all the serialize attributes.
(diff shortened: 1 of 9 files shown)


strum_macros/src/as_str.rs
```diff
@@ -0,0 +1,62 @@
...
+            Unit => quote::Ident::from(""),
+            Tuple(..) => quote::Ident::from("(..)"),
+            Struct(..) => quote::Ident::from("{..}"),
+        };
+
+        arms.push(quote!{ #name::#ident #params => #output });
+    }
+
+    if arms.len() < variants.len() {
+        arms.push(quote!{ _ => panic!("to_string() called on disabled variant.")})
```
Peternator7: Could you update this panic message for me?

lo48576:
Ok, I'll rewrite the code to use `derive(AsRefStr)` and common `to_string` attribute.

lo48576:
Removed `as_str` attribute and renamed `AsStr` to `AsRefStr`.

I'm sorry `to_string.rs` (https://github.com/Peternator7/strum/pull/10/files#diff-7eb81bb78680f1333200921a515022c5) has no content changes but trailing newline added...

Peternator7:
This is merged. I'll get a new version deployed on crates.io hopefully tonight :)

PR #11, opened 2017-10-24
Fix associated type in example generated code

wiiittttt:
Fix to match `type Err`


Peternator7:
@wiiittttt Thanks for the fix!! Merged :)

PR #14, opened 2018-02-08
don't suggest wildcard dependencies

durka:
As pointed out [on reddit](https://www.reddit.com/r/rust/comments/7vxmmy/macro_for_generating_string_enum_parser/dtvv8yt/).

You'll have to remember to update these when publishing new versions.


Peternator7:
Looks good to me. Merged

PR #25, opened 2018-05-23
Disable `missing_docs` lint for new types created by `EnumIter`

lo48576:
New structs generated by `derive(EnumIter)` caused `warning: missing documentation for a struct` warning on my environment (rustc 1.28.0-nightly (cb20f68d0 2018-05-21)).

I don't know whether this behavior (lint enabled for automatically generated code which users may not be responsible) will change later or not, but I think it is not so bad if these lints are automatically disabled for automatically generated codes.


Peternator7:
Makes sense!

PR #37, opened 2018-10-29
feature: add enum count macros

### LINKED ISSUE #15: Add EnumCount derive
I wrote this proc-macro because I needed it (for mapping Enum cases from different enums to Vec regions, making the Vec len equal to the combined number of all cases) and I think it would make sense to include it in this crate, because it already contains many useful derives for enums.
https://github.com/Boscop/enum-count/blob/master/src/lib.rs
https://github.com/Boscop/enum-count-derive/blob/master/src/lib.rs

It allows stuff like [this](https://github.com/Boscop/enum-count/blob/master/examples/basic.rs):
```rust
#[derive(EnumCount)] enum Aa { }

#[derive(EnumCount)]
enum Bb { B0 = AA_COUNT as _ }

#[repr(u8)]
#[derive(EnumCount)]
enum Cc { C0 = AA_COUNT as _, C1 = BB_COUNT as _ }

#[repr(usize)]
#[derive(EnumCount)]
enum Dd { D0 = AA_COUNT, D1 = BB_COUNT, D2 = CC_COUNT }

fn main() {
	println!("{} {} {} {}", Aa::count(), BB_COUNT, Cc::count(), DD_COUNT);
}
```

(The constant is necessary for it to be used in other enums and static array lens etc.)

Dushistov:
fix #15

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,6 +380,11 @@ Strum has implemented the following macros:
     }
     ```
 
+8. `EnumCount`: for a given enum generates implementation of `strum::EnumCount`,
+    which returns number of variants via `strum::EnumCount::count` method,
+	also for given `enum MyEnum` generates `const MYENUM_COUNT: usize`
+	which gives the same value as `strum::EnumCount` (which is usefull for array sizes, etc.).
+
 # Additional Attributes
 
 Strum supports several custom attributes to modify the generated code. At the enum level, the
--- a/strum/src/lib.rs
+++ b/strum/src/lib.rs
@@ -669,3 +669,8 @@ where
 {
     fn as_static(&self) -> &'static T;
 }
+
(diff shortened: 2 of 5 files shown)


Peternator7:
@Dushistov, thanks for the PR! This looks good. Would you be willing to write up the docs? If you can't, I'll work on them, but it will help get this deployed more quickly :)

Dushistov:
>  Would you be willing to write up the docs?

@Peternator7, I added description in README.md file

Peternator7:
Thanks for adding Docs!! I'll try and get this pushed to crates.io this week :)

Peternator7:
A little later than I would've liked, but the new version is on [crates.io](https://crates.io/crates/strum).

Thanks for the help :)

PR #40, opened 2018-12-20
implement From<Enum> for &'static str in case of AsStaticRef derive

### LINKED ISSUE #38: RFC: Use instead (in addition) to strum::AsStaticStr  core::convert::Into trait ?
What about generates something like:

```Rust
        impl Into<&'static str> for $Name {
            fn into(self) -> &'static str {
                match self {
                    $($Name::$Variant => $StrVar),*
                }
            }
        }
```

instead of `strum::AsStaticStr` or in additional to `strum::AsStaticStr`?

Dushistov:
fix #38

### DIFF
--- a/strum/src/lib.rs
+++ b/strum/src/lib.rs
@@ -372,7 +372,7 @@
 //!         assert_eq!(MyVariants::Variant0, MyEnum::Variant0(true).into());
 //!     }
 //!     ```
-//! 
+//!
 //! 8. `EnumCount`: for a given enum generates implementation of `strum::EnumCount`,
 //!    which returns number of variants via `strum::EnumCount::count` method,
 //!	also for given `enum MyEnum` generates `const MYENUM_COUNT: usize`
@@ -692,6 +692,7 @@ pub trait EnumProperty {
 
 /// A cheap reference-to-reference conversion. Used to convert a value to a
 /// reference value with `'static` lifetime within generic code.
+/// #[deprecated(since="0.13.0", note="please use `#[derive(IntoStaticStr)]` instead")]
 pub trait AsStaticRef<T>
 where
     T: ?Sized,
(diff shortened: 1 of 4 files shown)


Peternator7:
This is great! I'd say this is the way that it should've been done in the first place :) 

Here's my thoughts - when possible, I try and make it so someone can use `strum_macros` without `strum` itself. Rather than implement `From` in the same derive as AsStaticRef, can we make a separate custom derive that implements `From<Enum>` using `match x { (arms)* }`. That way there's no dependency on strum types and you can just import the macros :).

Then we'll work on deprecating `AsStaticRef`, but I'm not sure what the proper way of deprecating a procedural macro is.

Dushistov:
> Rather than implement From in the same derive as AsStaticRef, can we make a separate custom derive > that implements From<Enum> using match x { (arms)* }. That way there's no dependency on strum 
> types and you can just import the macros :).

I changed implementation as you suggested. I named new derive as `IntoStaticStr`,
yes it implements `From`  trait, but it purpose to convert into `&'static str`.

> Then we'll work on deprecating AsStaticRef, but I'm not sure what the proper way of deprecating a 
> procedural macro is.

I suppose it is enough to add deprecation note to trait (see my last comment) and user will see deprecation note.

Peternator7:
Merged!! Thanks for the PR. I'll work on getting this pushed to crates.

blacktemplar:
Why does `IntoStaticStr` deprecate `AsStaticRef`? What about the use case that I have a reference to an enum that does not implement `Copy` and I want to get a `&'static str` representation for it?

Peternator7:
`IntoStaticStr` should probably implement `From<&'a Enum> for &'static str` in addition to (or instead of) taking the enum by value so you can use it for types that are not copy.

PR #42, opened 2019-01-06
Fix serialize_all in `AsRefStr`, `AsStaticStr` and `IntoStaticStr`

xosdy:


Peternator7:
@xosdy , Thanks for opening a PR! Sorry for the delay in looking at it, I've been busy after the holidays. Could you confirm for me what this is fixing? Was serialize all not being invoked in specific instances?

xosdy:
Yes, `AsRefStr`, `AsStaticStr`, `IntoStaticStr` are not serialized by the `serialize_all` rule.

PR #48, opened 2019-04-09
Implement serialize_all=UPPERCASE

Dushistov:
I suppose it would be great if we would be compatible with serde:
https://serde.rs/container-attrs.html

> "lowercase", "UPPERCASE", "PascalCase", "camelCase", "snake_case", "SCREAMING_SNAKE_CASE", "kebab-case", "SCREAMING-KEBAB-CASE"

when it is clear from the name how text would be changed, like "camelCase" instead of "camel_case"

### DIFF
--- a/strum_macros/src/case_style.rs
+++ b/strum_macros/src/case_style.rs
@@ -7,18 +7,26 @@ pub enum CaseStyle {
     SnakeCase,
     TitleCase,
     UpperCase,
+    LowerCase,
+    ScreamingKebabCase,
+    PascalCase,
 }
 
 impl<'s> From<&'s str> for CaseStyle {
     fn from(text: &'s str) -> CaseStyle {
         match text {
-            "camel_case" => CaseStyle::CamelCase,
-            "kebab_case" => CaseStyle::KebabCase,
+            "lowercase" => CaseStyle::LowerCase,
+            "camel_case" | "PascalCase" => CaseStyle::PascalCase,
+            "kebab_case" | "kebab-case" => CaseStyle::KebabCase,
             "mixed_case" => CaseStyle::MixedCase,
-            "shouty_snake_case" | "shouty_snek_case" => CaseStyle::ShoutySnakeCase,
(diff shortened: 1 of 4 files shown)


Dushistov:
@Peternator7 

ping

Peternator7:
Thanks for the PR! I think matching Serde's naming is reasonable, but if we do it for one, then we should do it for all of them and just take the breaking change. Is that something you could do?

Dushistov:
@Peternator7 

> I think matching Serde's naming is reasonable,

Great!

> but if we do it for one, then we should do it for all of them and just take the breaking change. Is that something you could do?

I suppose there is no need for breaking changes, I don't see any intersections that have the same name for different cases. As first step we can just add new names:
```diff
- "camel_case" => CaseStyle::CamelCase,
+ "camel_case" | "camelCase" => CaseStyle::CamelCase,
```
in future may be there is sense to deprecate "camel_case" or may be not.

pitkley:
@Dushistov are you still working on this PR? I have a use-case for the `lowercase` renamer and would be happy to push this PR further.

Dushistov:
@pitkley 

> are you still working on this PR?

No, it just stuck on @Peternator7 's side,
in other words even if I plan to work on this I have no idea what to do to merge this PR.

Peternator7:
Sorry I didn't realize this was blocked on me. I'd like to get all the different cases implemented in a single pr if we can.

- [ ] snake_case
- [ ] screaming_snake_case
- [ ] kebab_case
- [ ] screaming_kebab_case
- [ ] lowercase
- [ ] uppercase
- [ ] camel_case
- [ ] pascal_case

I agree with @Dushistov that we don't need a breaking change to implement. Let's just add these in addition to the currently accepted values.

Dushistov:
> Let's just add these in addition to the currently accepted values

I add all from your list above (and from serde's doc, actually they are the same).
Little messy with 'camel_case', according to serde and google,
camel case starts with lower case, while pascal case starts with upper case,
but for compatibility I leave old name also,
so 'camel_case' and 'camelCase' are different.

Issue #53, opened 2019-05-23
Cannot use full-qualified path to trait inside `#[strum_discriminants(derive())]`


thomaseizinger:
When trying to use full-qualified paths for the custom derive, none of the attributes are applied.

For example:

```rust
#[allow(dead_code)]
#[derive(Debug, Eq, PartialEq, strum_macros::EnumDiscriminants)]
#[strum_discriminants(name(SplitAttributesBoo), derive(strum_macros::Display))]
#[strum_discriminants(derive(strum_macros::EnumIter))]
enum SplitAttributes {
    Variant0(bool),
    Variant1(i32),
}
```

I digged a bit into the code and it actually looks like `syn` doesn't report the full path in that case. I googled around a bit and found a similar problem here: https://github.com/rust-lang/rust/issues/55168

But not sure how relevant that actually is.
I created a PR that reproduces the issue here: #54.

PR #54, opened 2019-05-23
Add failing test for issue #53

thomaseizinger:


Peternator7:
It appears I need to fix the travis config as well. The test failed, but it didn't fail the build.

I'll poke around and see if there are any work arounds for this. Per the issue you posted, if it's an issue with how the compiler does name resolution, we should at least mention a workaround.

PR #57, opened 2019-06-23
Add feature gate `derive` for re-exporting proc macros to `strum`.

uHOOCCOOHu:
This is a shortcut to use proc macros without manually specifying `strum_macros` as a dependency, just as `serde` and many other packages do. So we can simply use `use strum::*;` (or `#[macro_use] extern crate strum;` for rustc prior to 1.31.0) to `derive`.

It is already tested to be compatible with rustc 1.26.0

Maybe more documentation is needed?

### DIFF
--- a/strum/Cargo.toml
+++ b/strum/Cargo.toml
@@ -12,8 +12,14 @@ documentation = "https://docs.rs/strum"
 homepage = "https://github.com/Peternator7/strum"
 readme = "../README.md"
 
+[dependencies]
+strum_macros = { path = "../strum_macros", optional = true, version = "0.15.0" }
+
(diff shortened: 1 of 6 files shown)


Dushistov:
This is great feature, now I have only

```
#[macro_use]
extern crate strum_macros;
```

on the top of my `lib.rs`/`main.rs`, all other `extern crate` I removed, thanks to 2018 edition.

@Peternator7 Is anything that should be fixed to make it possible to merge?

uHOOCCOOHu:
Rebased now

Peternator7:
This is great!! Sorry for the delay getting to it, but it'll be a great feature.

Issue #59, opened 2019-07-15
Reverse/DoubleEnded EnumIter


Nevsor:
It would be nice to be able to iterate over enums in reverse order.

Currently I have to use the workaround `my_enum::iter().collect::<Vec<_>>().iter().rev()`, which is quite ugly and probably less performant.

This could be implemented by making the IntoEnumIter::iter function return a DoubleEndedIterator to allow code like this:

``` rust
#[derive(EnumIter)]
enum Colors {Red, Green, Blue}

use strum::IntoEnumIterator;
for color in Colors::iter().rev() {
    // Do something with color.
}
```

This approach might incur a runtime cost, because we would have to add another field to the iterator struct (a second index for `next_back` similar to the existing `idx`) even if the user does not call `next_back`.

Another possible way to implement this is to add a new method, that returns a reverse or double-ended iterator.

Are you interested in adding this functionality?

Peternator7:
Thanks for the feedback! I think it's a very useful feature :) I'd be inclined to add a field to the existing iterator structure so as to keep the api surface small.

I'm unfortunately too busy to get to this in the short term, but if you open a PR, i'd be more than happy to review it.

Peternator7:
Thanks for the PR :) I'll get this published in `0.17`

PR #60, opened 2019-07-16
Implemented DoubleEndedIterator for EnumIter.

Nevsor:
This is a first working implementation of #59. 
All the tests pass including a new one to test the `DoubleEndedIterator`.

I have never used Rust macros before, so please tell me if something could be implemented better or something is missing.

Maybe it would be more elegant (and performant?) to generate a `static` or `const` array instead of the match arms and just return the arrays iterator in `iter`.


strum_macros/src/enum_iter.rs
```diff
@@ -79,9 +89,7 @@ pub fn enum_iter_inner(ast: &syn::DeriveInput) -> TokenStream {
             type Item = #name #ty_generics;
 
             fn next(&mut self) -> Option<#name #ty_generics> {
-                let output = match self.idx {
-                    #(#arms),*
-                };
+                let output = self.get(self.idx);
```
Peternator7: We actually need to add another check here. Iterating forward and backward shouldn't overlap so if `idx` is greater than `back_idx` we should return `None`

strum_macros/src/enum_iter.rs
```diff
@@ -62,14 +62,24 @@ pub fn enum_iter_inner(ast: &syn::DeriveInput) -> TokenStream {
...
+            }
+        }
+
         impl #impl_generics ::strum::IntoEnumIterator for #name #ty_generics #where_clause {
             type Iterator = #iter_name #ty_generics;
             fn iter() -> #iter_name #ty_generics {
                 #iter_name {
-                    idx:0,
+                    idx: 0,
+                    back_idx: 0,
```
Peternator7: I might set `back_idx` to `#variant_count - 1` so that one counts down and the other counts up.

strum_macros/src/enum_iter.rs
```diff
@@ -99,10 +107,24 @@ pub fn enum_iter_inner(ast: &syn::DeriveInput) -> TokenStream {
             }
         }
 
+        impl #impl_generics DoubleEndedIterator for #iter_name #ty_generics #where_clause {
+            fn next_back(&mut self) -> Option<#name #ty_generics> {
+                if self.back_idx >= #variant_count {
```
Peternator7: Same thing here, the iterators shouldn't overlap so if `back_idx` is less than `idx`, we are done.

strum_tests/tests/enum_iter.rs
```diff
@@ -100,3 +100,18 @@ fn cycle_test() {
     ];
     assert_eq!(expected, results);
 }
+
```
Peternator7: Can we get a test where we pull some from the front and some from the back so we ensure the non-overlapping behavior is handled correctly? Ideally one test should confirm that it works correctly if the last call is to `next()` and another should confirm it's correct if the last call is to `next_back()`

Peternator7:
Simply using an iterator to an array is an interesting option. It would require an additional `Clone` bound on the enum though.

Iterators to arrays return references to the items in the array. EnumIter returns values which means we would need to clone the values returned by the array's iterator to pass them back to the user, and that requires the type be `Clone`.

You can definitely look at that code to see how they avoid overlapping forward and backward iteration though!

strum_macros/src/enum_iter.rs
Nevsor: I tried this at first but got an underflow exception on the last element. This approach seemed like the easiest fix.

Nevsor:
Thank you for the feedback! I did not notice the intended overlapping behaviour on my first reading of the DoubleEndedIterator trait. Unfortunately I will be offline for the rest of the week and will not be able to fix this before tuesday.

Peternator7:
@Nevsor, are you still working on this?

strum_tests/tests/enum_iter.rs
Nevsor: Added the requested test and two others.

Nevsor:
@Peternator7, sorry for the delay. Here is my implementation of `DoubleEndedIterator`. I ended up implementing `next()` and `next_back()` as special cases of a new, more efficient implementation of `nth()` and `nth_back()`. 

dbc7a257b8cbe4aa98a38a7556fc3fb1fc398649 is not strictly related to this Pull Request, should it be merged seperately?

PR #61, opened 2019-08-13
Update attribute docs for strum_discriminants

tcharding:
Currently the README does not mention that extra attributes may be
passed to `#[strum_discriminants()]` and that these will be added as
attributes on the generated enum.  This is a cool feature, we should
document it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,6 +387,10 @@ fn main() {
 }
 ```
 
+When using `#[strum_discriminants()], the derive parameter may be followed by
+additional attributes and these will be expanded as attributes on the generated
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -387,6 +387,10 @@ fn main() {
 }
 ```
 
+When using `#[strum_discrimintants()], the derive parameter may be followed by
```
D4nte:
```suggestion
When using `#[strum_discriminants()], the derive parameter may be followed by
```

tcharding: Thanks, fixed and force pushed.

Peternator7:
Looks great! Thanks for adding some docs.

PR #65, opened 2019-08-22
Bump `proc-macro2`, `quote`, and `syn` to 1.0

lo48576:
* Add `helpers::eq_path_str` to compare a path and a string.
* Use `syn::Attribute::parse_meta()` instead of `interpret_meta()`.
* Extract ident from path, if necessary.
    + Some of `syn` types now have `path` member instead of `ident`.

Closes #63.

PR #66, opened 2019-08-22
Refactoring (mainly fixing clippy lint warnings)

lo48576:
NOTE: This branch is based on #65.

Fixing clippy lint warnings, and some tiny changes which does not change strum's behavior.

I'm wondering whether the commit <https://github.com/Peternator7/strum/commit/ff4f6087e4a4f7a5fd0834871a7d928d13721866> should be included in this pull request.
This uses `Iterator::copied()` which is stabilized since Rust 1.36, but I'm not sure what version of Rust strum should support.
When strum use `syn`-1.0, it bumps minimal Rust requirement version to Rust 1.31 or later.
Can it be Rust 1.36?

### DIFF
--- a/strum/src/lib.rs
+++ b/strum/src/lib.rs
@@ -59,7 +59,7 @@ impl std::fmt::Display for ParseError {
         // We could use our macro here, but this way we don't take a dependency on the
         // macros crate.
         match self {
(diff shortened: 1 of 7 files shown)


lo48576:
Outdated.

PR #68, opened 2019-08-26
Fix inaccurate case_style

DrSensor:
especially on `#[strum(serialize_all = camel_case)]`


Dushistov:
It is not "inaccurate", see https://github.com/Peternator7/strum/pull/48#issuecomment-515269369 new camelCase and old "camel_case" doesn't match because of compatibility, new behavior make break someone code.

DrSensor:
Ah, how about renaming the panic messages to follow [`#[serde(rename_all = ...)]`](https://serde.rs/container-attrs.html#rename_all) so it can be soft deprecated? 

(I will revert back the `"camel_case"` to point `CaseStyle::PascalCase`)

PR #74, opened 2019-11-02
Move EnumVariantsNames into a trait which the macro implements

Spanfile:
I ran into a case where I had to accept a generic parameter of any enum type that has the `variants()` function implemented, which would've previously required some awkward workarounds. This PR moves the function into its own trait `VariantNames` and changes the `EnumVariantNames` to implement it, so that it is possible to use the trait to accept any such enum (see the added test case).

### DIFF
--- a/strum_macros/src/macros/enum_variant_names.rs
+++ b/strum_macros/src/macros/enum_variant_names.rs
@@ -1,8 +1,7 @@
 use proc_macro2::TokenStream;
 use syn;
 
-use crate::helpers::case_style::CaseStyle;
-use crate::helpers::{extract_meta, CaseStyleHelpers, MetaIteratorHelpers};
+use crate::helpers::{case_style::CaseStyle, extract_meta, CaseStyleHelpers, MetaIteratorHelpers};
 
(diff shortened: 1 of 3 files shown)


strum_macros/src/macros/enum_variant_names.rs
```diff
@@ -24,10 +23,10 @@ pub fn enum_variant_names_inner(ast: &syn::DeriveInput) -> TokenStream {
         .collect::<Vec<_>>();
 
     quote! {
-        impl #name {
+        impl VariantNames for #name {
```
Peternator7: We have to do a little work here to match types with generic parameters. See the [EnumIter macro](https://github.com/Peternator7/strum/blob/e981f240d6e02c49f246b0fc4c7f4978e5ff364b/strum_macros/src/macros/enum_iter.rs#L68) for an example. I suspect the old implementation missed this so it would be great if you could fix it.

Peternator7: Also make sure we use an absolute path to the trait to avoid collisions

Peternator7:
Looks good! Merged

PR #75, opened 2019-11-06
Const enum variants names

WaffleLapkin:
I tried to write code like this:
```rust
pub const LETTERS: &'static [&'static str] = Letter::variants();

#[derive(EnumVariantNames)]
pub enum Letter { A, B, C }
```
But I've found that `variants` fn is not `const`.

This pr changes the behaviour of `EnumVariantNames` in that way:
1. Add `pub const VARIANTS: &'static [&'static str] = ...` to enum
2. make `variants` fn `const`
3. deprecate `variants` fn with a message like ``"Use `Letter::VARIANTS` instead"``

### DIFF
--- a/strum/src/lib.rs
+++ b/strum/src/lib.rs
@@ -223,11 +223,6 @@ pub trait EnumCount {
 pub trait VariantNames {
     /// Names of the variants of this enum
     const VARIANTS: &'static [&'static str];
-
-    /// Return a slice containing the names of the variants of this enum
-    fn variants() -> &'static [&'static str] {
-        Self::VARIANTS
-    }
 }
 
 #[cfg(feature = "derive")]
--- a/strum/src/lib.rs
+++ b/strum/src/lib.rs
@@ -221,7 +221,13 @@ pub trait EnumCount {
 /// A trait for retrieving the names of each variant in Enum. This trait can
 /// be autoderived by `strum_macros`.
 pub trait VariantNames {
-    fn variants() -> &'static [&'static str];
+    /// Names of the variants of this enum
+    const VARIANTS: &'static [&'static str];
+
+    /// Return a slice containing the names of the variants of this enum
+    fn variants() -> &'static [&'static str] {
+        Self::VARIANTS
+    }
 }
 
 #[cfg(feature = "derive")]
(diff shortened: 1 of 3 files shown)


hobofan:
> deprecate variants fn with a message like "Use `Letter::VARIANTS` instead"

Why deprecate the function if you made the access via the function const too?

WaffleLapkin:
Because I see no reasons to use the function, when you can use associated constant.

Peternator7:
Thanks for the PR! Sorry I thought I responded to this last week, but it looks like I forgot to. Based on #74, I don't think we should deprecate the function because it will be extracted into a trait, and those can't be `const` yet.

Let's definitely get the associated constant added though :)

WaffleLapkin:
> Based on #74, I don't think we should deprecate the function because it will be extracted into a trait, and those can't be const yet.

So I must undeprecate `variants` and make it non-`const` again? Ok, I'll make these changes tomorrow :)

Peternator7:
Thanks! Merging #74 created a conflict so here's the resolution you'll want in the enum variants file. 

```rust
    quote! {
        impl #impl_generics ::strum::VariantNames for #name #ty_generics #where_clause {
            /// Return a slice containing the names of the variants of this enum
            #[allow(dead_code)]
            fn variants() -> &'static [&'static str] {
                Self::VARIANTS
            }
        }

        impl #impl_generics #name #ty_generics #where_clause {
            /// Names of the variants of this enum
            #[allow(dead_code)]
            pub const VARIANTS: &'static [&'static str] = &[ #(#names),* ];
        }
    }
```

WaffleLapkin:
@Peternator7 shouldn't I add `VARIANTS` const to the trait?

Peternator7:
@WaffleLapkin, that would be preferred, but if I remember correctly, const in traits isn't supported on stable yet :(

WaffleLapkin:
@Peternator7 `const` in traits [is supported] on stable

[is supported]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b6ac0c13065e798fbc51a43378ab649b

WaffleLapkin:
@Peternator7 I've added `VARIANTS` const to `VariantNames` trait and it seems like there is still no need in `variants` function... Am I wrong?

WaffleLapkin:
@Peternator7 ping

Peternator7:
Thanks for the ping; I agree. Let's go ahead and remove the variants function

Peternator7:
Looks great! Thanks for the feature :)

WaffleLapkin:
Yay! Btw, don't forget that these changes are breaking now ;)

PR #76, opened 2019-11-08
Add checks to guard against overflow in iterator methods.

hermanventer:
When a generated iterator runs of the end it keeps on incrementing the index variable, which is undesirable. Furthermore the size_hint method does not guard against the index being greater than the variant count, which causes an overflow.

These issues cause our static analyses [tool](https://github.com/facebookexperimental/MIRAI) to generate warnings when analyzing code generated by strum, which is annoying.

Thanks for considering this.

### DIFF
--- a/strum_macros/src/macros/enum_iter.rs
+++ b/strum_macros/src/macros/enum_iter.rs
@@ -83,12 +83,14 @@ pub fn enum_iter_inner(ast: &syn::DeriveInput) -> TokenStream {
                     #(#arms),*
                 };
 
-                self.idx += 1;
+                if self.idx < #variant_count {
+                    self.idx += 1;
+                }
                 output
             }
 
             fn size_hint(&self) -> (usize, Option<usize>) {
-                let t = #variant_count - self.idx;
+                let t = if self.idx >= #variant_count { 0 } else { #variant_count - self.idx };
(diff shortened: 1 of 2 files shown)


Peternator7:
Looks great! Thanks for the PR :)

strum_macros/src/macros/enum_iter.rs
```diff
@@ -83,12 +83,14 @@ pub fn enum_iter_inner(ast: &syn::DeriveInput) -> TokenStream {
...
-                self.idx += 1;
+                if self.idx < #variant_count {
+                    self.idx += 1;
+                }
                 output
             }
 
             fn size_hint(&self) -> (usize, Option<usize>) {
-                let t = #variant_count - self.idx;
+                let t = if self.idx >= #variant_count { 0 } else { #variant_count - self.idx };
```
Ten0, line 93: Looks like this is unnecessary given that the above guarantees this will never underflow.

Additionally, this property is [`FusedIterator`](https://doc.rust-lang.org/std/iter/trait.FusedIterator.html) (which is not a requirement for `Iterator`, and it's the responsibility of the caller to stop calling it). If we choose to have it, then we should also signify that we have it by implementing `FusedIterator`.

PR #79, opened 2019-11-29
Replace 'variantes' with 'variants' in README.MD

cedric-h:


Peternator7:
Looks great! Thanks for the fix :)

PR #84, opened 2020-02-07
Apply serialize_all casing to get_serializations

t-mw:
I'm not sure if this is intentional, but it seems to make sense to convert case for `get_serializations`. When `to_string` is already being used for display you may still want to use the automatic case conversion for serialization.


Peternator7:
I'd agree it seems more consistent. Would you mind adding a note to the changelog as well?

t-mw:
Ok great. Sure, I've added the note.

Peternator7:
Merged! Thanks for the fix :)

PR #89, opened 2020-03-22
Extract strum_macros::helpers into it's own module

ethankhall:
This was done so I could access the helpers in another project.


ethankhall:
@Peternator7 any chance you'll be able to take a look at this sometime?

Peternator7:
Hi @ethankhall, sorry for the delay here. The whole Covid thing has really messed up my schedule, and this got lost somewhere. Thanks for the ping :) 

Are you writing your own proc-macros? I'm not opposed to refactoring this code into it's own package, but I'd like to understand your use-case a little better to make sure this is really the best way of doing it.

Thanks!

ethankhall:
No worries! Stuff falls through the cracks from time to time.

Sure, I can actually share what I'm doing with you. The goal was to make it easier to make CLI's with errors that generate error codes on failure.

Here's how I've been using my branch this change: https://github.com/ethankhall/error-enum-generator

strum_helpers/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+[lib]
+name = "strum_helpers"
+
+[dependencies]
+heck = "0.3"
+proc-macro2 = "1.0"
+syn = { version = "1.0", features = ["parsing", "extra-traits"] }
+
+[features]
+verbose-enumstring-name = []
```
Peternator7, line 24: Do we need these feature flags on the helpers crate? I think they are just needed in the primary "macros" crate.

ethankhall: Sorry life's been super busy. I don't think so, it was mostly copy-paste. I'll update the branch & pr here shortly.

PR #105, opened 2020-08-06
Fixes a panic caused by unwrapping an option

Peternator7:


strum_macros/src/helpers/has_metadata.rs
```diff
@@ -11,7 +11,7 @@ fn get_metadata_inner<'a>(
     it: impl IntoIterator<Item = &'a syn::Attribute>,
 ) -> Vec<syn::Meta> {
     it.into_iter()
-        .map(|attr| attr.parse_meta().unwrap())
+        .filter_map(|attr| attr.parse_meta().ok())
```
jplatte, line 14: AFAICT this is still wrong. This will now silently ignore malformed `strum` attributes. Instead, you should filter by attr.[path](https://docs.rs/syn/1.0.40/syn/struct.Attribute.html#structfield.path).is_ident(ident) and then use `parse_meta()` without discarding errors (the `is_ident` check below can then be skipped).

jplatte: I can also recommend setting up [trybuild](https://github.com/dtolnay/trybuild) to test expected proc-macro errors.

Peternator7: Happy to take a PR fixing this. Historically strum ignored improperly formatted/tagged attributes. Version `0.19` took some steps in the right direction in terms of failing on bad inputs, but not all of them.

Issue #77, opened 2019-11-21
Naming problems


frol:
I wonder how you manage the non-descriptive names like `as_static`, `as_ref`, when you have an enum that, e.g. describes a state:

```rust
enum MyState {
    Position1(i32),
    GameOver(String),
}
```

Having `my_state.as_ref()` or `my_state.as_static()` looks odd to me when I really want to express `my_state.as_variant_name()`. Currently, I have to write a 4-line wrapper around `as_static` just to make my API sane, but I still have a confusing `AsStaticRef` implementation in the documentation.

This is a small papercut, but I still want to bring it in.

I would prefer having `AsEnumVariantName` trait with `as_enum_variant_name` or `as_variant_name` method.

Peternator7:
Sorry I didn't respond sooner; I've been thinking about this comment for a while now, and I tend to agree, I'm just looking for a good time to make the change because it will be rather disruptive I'm sure.

frol:
May I suggest introducing new traits and method names while keeping the old ones as aliases? So then the old names can be promoted to the deprecation and the future breaking-change release would just remove the old names altogether.

PR #106, opened 2020-08-06
Add a Derive for VariantName which provides a variant_name method

nipunn1313:
This is intentionally different from
`impl From<Enum> for &'static str {...}`

This intends to be more declarative about the meaning of the string.

There are many ways to convert an enum to a string (eg display, debug,
variant name, etc), and as_static() and into() are ambiguous and feel
overloaded for some callers.

Some callers may want to have a more declarative unambiguous name for
a method with this signature fn (&self) -> &'static str.

This provides it! Implementation wise, it's identical to the
deprecated as_static()


.gitignore
```diff
@@ -1,2 +1,3 @@
 target
 Cargo.lock
+.*.swp
```
Peternator7, line 3: What is this?

Peternator7:
Hi @nipunn1313, 

Thank you for the PR. I like the idea of this PR, but I'm not ready to merge it yet.

As I'm sure you've seen, there's a proliferation of string methods in strum, and part of the difficulty is getting the signatures right and choosing good names for things. While I think `variant_name` is a reasonable choice, I'm sure someone will want to name the method something else for their own specific domain, and I'm not sure what the best way for us to handle that is yet.

Also see #77

.gitignore
nipunn1313: Ah it was for vim swap files - was gitignoring it.

PR #110, opened 2020-09-02
Update strum to Rust edition 2018

### LINKED ISSUE #109: Convert to Rust edition 2018?
According to `README.md`, the MSRV for strum is 1.31, which is the version that introduced the 2018 edition. Would a PR updating the code base to the new edition be welcome?

jplatte:
Fixes #109.

I've updated the docs too (removing the mentions of `extern crate`), is that okay?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,18 +23,6 @@ strum = "0.19"
 strum_macros = "0.19"
 ```
 
-And add these lines to the root of your project, either lib.rs or main.rs.
-
-```rust
-// Strum contains all the trait definitions
-extern crate strum;
-#[macro_use]
-extern crate strum_macros;
-
-// Instead of #[macro_use], newer versions of rust should prefer
-use strum_macros::{Display, EnumIter}; // etc.
-```
-
 # Strum Macros
 
 Strum has implemented the following macros:
--- a/strum/Cargo.toml
+++ b/strum/Cargo.toml
@@ -1,6 +1,7 @@
 [package]
 name = "strum"
 version = "0.19.2"
+edition = "2018"
 authors = ["Peter Glotfelty <peter.glotfelty@microsoft.com>"]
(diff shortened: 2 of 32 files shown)


Peternator7:
Thanks for opening the PR :) It looks like the build failures are legit, and there's probably some missing usings in the test cases. I'll be good to merge this once we fix the build.

jplatte:
Huh, so really the way to solve this would be changing `#[cfg(feature = "derive")]` on the macro re-exports to `#[cfg(any(feature = "derive", doctest))]`, but that [doesn't work](https://github.com/rust-lang/rust/issues/67295). Alternatively CI could only run tests with `--all-features`, but that seems like a pretty bad idea. WDYT?

jplatte:
@Peternator7 friendly ping.

Peternator7:
I was poking around with this locally, and it looks to me like it works if I call `cargo test` from the root of the repository (the workspace) but fails if it's run from any of the subprojects. Can you confirm whether that behavior works for you?

If that's the case, we might be able to just update the appveyor script to call `cargo test` from the repo root instead of needing to step in and out of each of the projects.

jplatte:
Yes, by calling `cargo test` from the workspace root the strum derive feature is enabled because it is required by another workspace member. `cargo test --features derive` or `cargo test --all-features` inside the strum directory should work too. Do your okay with doctests boot working without that feature turned on?

Peternator7:
So I think I'm okay with that. The only people really running the tests should be contributors to this repo, and I would expect the most natural thing for someone to do is build the workspace and that works as expected. I pushed some changes last night that fixed the build and started actively building 1.31 for appveyor.

The gates on this PR pass now so let me review the content, but we should be good wrt the build.

Peternator7 (approved):

PR #119, opened 2020-09-29
Validate strum attributes

jplatte:
Fixes https://github.com/Peternator7/strum/pull/105#discussion_r487763261.


Peternator7:
Looks good! Thanks for the fix

Issue #115, opened 2020-09-22
Why are the macros documented as a GitHub wiki?


jplatte:
This means that you can't just open a PR improving (this part of) the docs, or easily access the docs for an older version. I think there were some issues with documenting proc macros in rustdoc in the past, but they have long been fixed. Could the docs be moved out of the Wiki?

Peternator7:
The annoying thing about documenting them with rust-docs was that I needed to publish a new version to just update the docs which slowed down the process. I'm not opposed to switching them back, but it's not a particularly high priority.

mike-kfed:
would you accept a pull-request for moving docs to the source? Because e.g. the wiki entry for EnumVariantNames does not match reality. In the first paragraph it references the variants() method, which no longer exists. The clap example code uses the constant but does not import from strum_macros. The structopt example is using old structopt raw and the old variants() method. If those code examples were doctests they would be in working order, right now none of those even compile. Other option: they could be example code, in the examples folder.
Anyways I'd be happy to help with any of that, as this will make it easier for new users starting to use this crate.

Peternator7:
Hi @mike-kfed, yes I'd be happy to accept a PR if you put one together :)

mike-kfed:
putting it in the strum docs is not as easy as expected, looking at rust-lang/cargo#2760 documenting code from examples is not included in the normal docs. This allows following options:

1. we tell people in the docs to check the examples folder where a README explains the examples.
1. we use a similar script as f3 (see cargo-doc issue) to create a "examples module" to have code in the html docs
1. we actually put code docs into the strum_macros crate and only additional examples in examples folder

p.s. I started with option 1 first, see here https://github.com/mike-kfed/strum/tree/wiki_to_examples/strum/examples as this will be the most straight forward, given that the examples are not really tests. `cargo build --examples` shows at least if they compile ;)

please let me know which option you prefer and I'll continue working on this :)

jplatte:
> we tell people in the docs to check the examples folder where a README explains the examples.

This still requires jumping around, even more if you want to look at the examples for an older version :slightly_frowning_face: 

I'd much prefer doctests (3).

mike-kfed:
doctests would work fine except for the structopt example. To allow for any example to be in the docs we'd need to run the copy script that f3 does (which creates an examples module in src/ that only holds documentation) = solution A.
solution B - rewrite the examples to be doc-tests (=remove the structopt example*).

[...]

PR #121, opened 2020-10-05
Wiki to doc tests

mike-kfed:
second attempt to fix issue #115 

I did not include the example using structopt since that does not compile on your CI tests.

the link to the attributes page of the wiki remains. didn't figure out where to put this in the code (maybe extra rust source file that is just documentation?)

### DIFF
--- a/strum/src/additional_attributes.rs
+++ b/strum/src/additional_attributes.rs
@@ -1,0 +1,68 @@
+//! # Documentation for Additional Attributes
+//!
+//! Strum supports several custom attributes to modify the generated code. At the enum level, the
+//! `#[strum(serialize_all = "snake_case")]` attribute can be used to change the case used when
+//! serializing to and deserializing from strings:
+//!
+//! ```rust
+//! use std::string::ToString;
+//! use strum;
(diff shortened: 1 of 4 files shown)


jplatte:
> the link to the attributes page of the wiki remains. didn't figure out where to put this in the code (maybe extra rust source file that is just documentation?)

I think what you could do for attributes is have one `.md` file per attribute that describes that attribute, and add

```rust
///
/// # Attributes
///
#[doc(include("attr_foo.md"))]
#[doc(include("attr_bat.md"))]
```

at the end of the docs for every drive macro that supports a set of attributes.

mike-kfed:
while a good idea, this is experimental see rust-lang/rust#44732

I'd rather wait for this to be stabilised.

jplatte:
Oh, I thought that was stable. In that case I might make a PR adding sth like this behind a feature flag later but that probably shouldn't block the current work.

Peternator7:
This is super great! Thanks for taking on this work. I'll merge it and push a patch version to crates today so docs.rs picks up the new version.

PR #134, opened 2020-10-11
More attribute refactoring

jplatte:
This has two effects that may not be obvious from the commit messages / on first sight:

* `strum(default = "true")` and `strum(disabled = "true")` no longer produce a dedicated "deprecation" compiler error
* the `strum_discriminants` "passthrough" of other attributes now supports attributes that don't parse as `syn::Meta`, e.g. `foo(key = some::Path)`, since syn only supports literals in the value position of `MetaList`s but the compiler allows arbitrary token streams in attributes

I'll probably work on converting all of the panics to `syn::Error`s next :)


strum_macros/src/helpers/case_style.rs
```diff
@@ -1,5 +1,9 @@
 use heck::{CamelCase, KebabCase, MixedCase, ShoutySnakeCase, SnakeCase, TitleCase};
-use syn::Ident;
+use std::convert::TryFrom;
```
Peternator7, line 2: Unfortunately, `TryFrom` isn't available until `1.34`. It annoys me, but I'm not so annoyed that I'm willing to up the MSRV yet.

strum_macros/src/helpers/metadata.rs
```diff
@@ -0,0 +1,241 @@
+use proc_macro2::{Span, TokenStream};
```
Peternator7, line 1: This whole file is very cool I have clearly not been keeping up with syn's features :)

strum_macros/src/helpers/type_props.rs
```diff
@@ -1,87 +1,56 @@
...
 }
 
-#[derive(Debug, Clone, Eq, PartialEq, Default)]
+#[derive(Debug, Clone, Default)]
 pub struct StrumTypeProperties {
     pub case_style: Option<CaseStyle>,
     pub discriminant_derives: Vec<Path>,
-    pub discriminant_name: Option<Path>,
-    pub discriminant_others: Vec<Meta>,
+    pub discriminant_name: Option<Ident>,
```
Peternator7, line 17: I'm trying to think of any places where a path would be valid here, but I don't think it ever is so this is good.

strum_macros/src/macros/enum_discriminants.rs
```diff
@@ -130,7 +126,7 @@ pub fn enum_discriminants_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
     Ok(quote! {
         /// Auto-generated discriminant enum variants
         #derives
-        #(#pass_though_attributes)*
+        #(#[ #pass_though_attributes ])*
```
Peternator7, line 129: Could you help me understand the difference here?

Peternator7 (changes requested):
Looks really great! Thanks for putting all this work in. Couple of minor points, and the `TryFrom` back-compat issue, but otherwise not much.

strum_macros/src/helpers/case_style.rs
jplatte: Haha, hadn't looked at the MSRV. I'll just switch from `TryFrom` implementation to constructor fn, does that sound good?

strum_macros/src/macros/enum_discriminants.rs
jplatte: `#()` interpolations can contain arbitrary extra tokens in addition to an iterable of `impl ToTokens`. So instead of 

```rust
let values = vec![1, 2, 3];
let push_values: Vec<_> = values.into_iter().map(|v| quote! { vec.push(#v); }).collect();
quote! {
    let vec = Vec::new();
    #(#push_values)
}
```

you can just write

```rust
let values = vec![1, 2, 3];
quote! {
    let vec = Vec::new();
    #( vec.push(#values); )
}
```

I did the same kind of simplification here ([context](https://github.com/Peternator7/strum/pull/134/files/15596f68516f2e6728fb7cd59507f3bb63482d21#diff-189e994b49682e22ee918c0157e1638eR40-R41)).

strum_macros/src/helpers/case_style.rs
jplatte: Actually, can make it a `FromStr` implementation!

Peternator7:
I don't think `Self` alias for enums was stablized until [1.37](https://github.com/rust-lang/rust/issues/49683). Otherwise, still seems good.

jplatte:
Sorry, not getting notifications about Travis CI failure. Should be fixed now :)

Peternator7 (approved):

Peternator7:
Everything looks good. Thanks for all the changes! Do you have more coming?

jplatte:
Yeah, still planning to replace `panic!`s by `syn::Error` and implement most or all of the features I have recently requested. Can't make any promises as for when that'll happen, but I hope to have another PR up in the coming days / weeks.

PR #138, opened 2020-11-02
Adding visibility option `vis(...)` to `#[strum_discriminants]` (#137)

### LINKED ISSUE #137: Add ability to specify visibility in generated type via `#[strum_discriminants]`
By "visibility", I'm referring to `pub`, `pub(crate)`, etc. I have an enum that I want to keep crate-scoped, but I'd like to generate discriminants and have those discriminants be exported publicly. Perhaps it could be another property in `#[strum_discriminants]`, such as: `#[strum_discriminants(vis(pub))]`?

linclelinkpart5:
Closes #137 

Validated using doctest example in the `EnumDiscriminants` section. While I was able to show in the doctest that I'm able to use the public discriminants properly (i.e. they work and are visible), I'm not sure how to show that I'm *not* exposing the original enum publicly. If anyone has suggestions for that, that would be helpful to include!

### DIFF
--- a/strum_tests/tests/enum_discriminants.rs
+++ b/strum_tests/tests/enum_discriminants.rs
@@ -215,3 +215,43 @@ fn filter_variant_attributes_pass_through() {
         VariantFilterAttrDiscs::from_str("dark_black").unwrap()
     );
 }
+
+#[cfg(bare_pub)]
+#[test]
+fn override_visibility() {
+    mod private {
+        use super::*;
+
+        #[allow(dead_code)]
+        #[derive(EnumDiscriminants)]
+        #[strum_discriminants(name(PubDiscriminants), vis(pub))]
+        enum PrivateEnum {
+            VariantA(bool),
+            VariantB(bool),
+        }
+    }
+
+    use private::PubDiscriminants;
+
+    assert_ne!(PubDiscriminants::VariantA, PubDiscriminants::VariantB);
+}
+
+#[cfg(not(bare_pub))]
+#[test]
+fn override_visibility() {
+    mod private {
+        use super::*;
+
+        #[allow(dead_code)]
+        #[derive(EnumDiscriminants)]
+        #[strum_discriminants(name(PubDiscriminants), vis(r#pub))]
+        enum PrivateEnum {
+            VariantA(bool),
+            VariantB(bool),
+        }
+    }
+
+    use private::PubDiscriminants;
+
+    assert_ne!(PubDiscriminants::VariantA, PubDiscriminants::VariantB);
+}
--- a/strum_tests/tests/enum_discriminants.rs
+++ b/strum_tests/tests/enum_discriminants.rs
@@ -231,9 +231,10 @@ fn override_visibility() {
         }
     }
 
-    use private::PubDiscriminants;
-
-    assert_ne!(PubDiscriminants::VariantA, PubDiscriminants::VariantB);
+    assert_ne!(
+        private::PubDiscriminants::VariantA,
+        private::PubDiscriminants::VariantB,
+    );
 }
 
 #[cfg(not(bare_pub))]
@@ -251,7 +252,8 @@ fn override_visibility() {
         }
     }
 
-    use self::private::PubDiscriminants;
-
-    assert_ne!(PubDiscriminants::VariantA, PubDiscriminants::VariantB);
+    assert_ne!(
+        private::PubDiscriminants::VariantA,
+        private::PubDiscriminants::VariantB,
+    );
 }
--- a/strum_tests/tests/enum_discriminants.rs
+++ b/strum_tests/tests/enum_discriminants.rs
@@ -251,7 +251,7 @@ fn override_visibility() {
         }
     }
 
-    use private::PubDiscriminants;
+    use self::private::PubDiscriminants;
 
     assert_ne!(PubDiscriminants::VariantA, PubDiscriminants::VariantB);
 }
--- a/strum_macros/src/helpers/metadata.rs
+++ b/strum_macros/src/helpers/metadata.rs
@@ -4,7 +4,7 @@ use syn::{
     parse::{Parse, ParseStream},
     punctuated::Punctuated,
     spanned::Spanned,
-    Attribute, DeriveInput, Ident, LitStr, Path, Token, Variant,
+    Attribute, DeriveInput, Ident, LitStr, Path, Token, Variant, Visibility,
 };
 
 use super::case_style::CaseStyle;
@@ -18,6 +18,7 @@ pub mod kw {
     // enum discriminant metadata
     custom_keyword!(derive);
     custom_keyword!(name);
+    custom_keyword!(vis);
 
     // variant metadata
     custom_keyword!(message);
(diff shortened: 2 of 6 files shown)


linclelinkpart5:
Seems that Rust 1.31 isn't happy with using `pub` in the macro attributes, but the other Rust versions are. This would be outside the scope of my current knowledge of proc macros, is there a way this could be made to work with 1.31? And curious what Rust version this became legal?

Peternator7:
Thanks for jumping in to help!! 

I haven't seen an error like this before, but jumping around rust versions, it appears this behavior changed between 1.33 and 1.34, and it looks like the error is with rustc parsing the "pub" in the macro attribute as a keyword, and not actually related to the code being generated.

linclelinkpart5:
If strict compatibility with Rust 1.31+ is needed, perhaps the `vis` keyword and logic could be put behind a feature gate?

Peternator7:
I'm not too worried about compatibility because this is a new feature so it **shouldn't** break any existing code.

Let's do this:
- Try using `vis(r#pub))` in the doc? I believe `r#keyword` let's you escape it, and that might fix the parse error in older versions of rust.
- If that works, let's add a comment above it that says on versions of rust 1.34 and greater, the r# is unnecessary.
- Can you add a test to the tests subproject? The docs get refactored more frequently than the tests so I'd like to make sure it's in a "permanent" spot. All's the test really needs to do is make sure `vis(pub)` and `vis(pub(crate))` compile. Like you've said elsewhere. There's not a lot of functionality to validate.
- If getting `vis(pub(crate))` to compile on 1.31 is too difficult, you can add a build cfg to the tests crate to skip that test. Here's the [`build.rs`](https://github.com/Peternator7/strum/blob/25eaa5a12c4b8a2530fef0c630ff45f2c00951bd/strum_tests/build.rs) and [here's an  example][1] of using the custom cfg.

[1]: https://github.com/Peternator7/strum/blob/1d2778842b8e859d1367712b881da695b2ca0c67/strum_tests/tests/enum_discriminants.rs#L185

linclelinkpart5:
@Peternator7 Thanks! I tried out using "r#pub", and now it's the opposite: it only works on 1.31, and breaks the other Rust version builds. It seems I'd need to do a cfg of some sort for Rust <= 1.33 to use `r#pub`, and just `pub` otherwise.

linclelinkpart5:
And to make things even trickier, it seems that even behind a `#[cfg(not(bare_pub))]` guard (where `bare_pub` is set for Rust >= 1.34), `rustc 1.31` still sees the `pub` keyword in the other `#[cfg(bare_pub)]` and fails. I'm not sure what to do here, it seems that nothing is working. :disappointed:

Peternator7:
Quite the catch-22 isn't it 👎 I think I see a few options here:

1. Change the syntax to for visibility monitors to use a different word than `pub`. This seems unideal and requires more docs to explain the feature.
2. Up the MSRV to 1.34. This will probably happen eventually, but I don't think it's worth doing right now.
3. Split this feature into a separate code example and add a `nocompile` to the doc so it just doesn't get compiled.

I'm leaning towards option 3 because the other 2 seem unfeasible. It's not great that the doc won't be tested, but you've added test cases so we should still catch any regressions.

Peternator7:
When I pulled your branch locally, I couldn't repro the `bare_pub` issue you were describing though so I'm less sure what the issue is there.

linclelinkpart5:
@Peternator7 Of the three, I agree that no. 3 sounds the best. I'll proceed down that route, this would be a separate code block example right underneath the main one, correct?

> When I pulled your branch locally, I couldn't repro the `bare_pub` issue you were describing though so I'm less sure what the issue is there.

I saw this error on travis, trying to build using Rust 1.31. I didn't try changing versions of my Rust toolchain locally, it was using Rust 1.47.

linclelinkpart5:
Suggestion no. 3 worked great! Travis and AppVeyor look happy with the results. :smile:

strum_tests/tests/enum_discriminants.rs
```diff
@@ -215,3 +215,45 @@ fn filter_variant_attributes_pass_through() {
         VariantFilterAttrDiscs::from_str("dark_black").unwrap()
     );
 }
+
+#[cfg(bare_pub)]
+#[test]
```
Peternator7, line 220: One more thing, can you use `cfg_attr` on the `strum_discriminants` attribute to determine between `pub` and `r#pub` instead of duplicating the entire function?

```rust
#[test]
fn override_visibility() {
    mod private {
        // ...

        #[allow(dead_code)]
        #[derive(EnumDiscriminants)]
        #[strum_discriminants(name(PubDiscriminants))]
        #[cfg_attr(bare_pub, strum_discriminants(vis(pub)))]
        #[cfg_attr(not(bare_pub), strum_discriminants(vis(r#pub)))]
        enum PrivateEnum {
            VariantA(bool),
            VariantB(bool),
        }
    }

    // ...
}
```

Peternator7 (changes requested):
Looks great! Glad the builds are working. Only one comment. I think we can deduplicate some code.

strum_tests/tests/enum_discriminants.rs
linclelinkpart5: Makes sense, I'll make the change now!

linclelinkpart5: Hmmm, looks like my Rust 1.47 doesn't like seeing `r#pub` at all (file is `strum_tests/tests/enum_discriminants.rs`):
```
mark@rche:~/Projects/rust/strum$ cargo test
   Compiling strum_tests v0.19.4 (/home/mark/Projects/rust/strum/strum_tests)
error: unexpected token
   --> strum_tests/tests/enum_discriminants.rs:228:59
    |
228 |         #[cfg_attr(not(bare_pub), strum_discriminants(vis(r#pub)))]
    |                                                           ^^^^^

error[E0433]: failed to resolve: could not find `PubDiscriminants` in `private`
   --> strum_tests/tests/enum_discriminants.rs:236:18
    |
236 |         private::PubDiscriminants::VariantA,
    |                  ^^^^^^^^^^^^^^^^ could not find `PubDiscriminants` in `private`

error[E0433]: failed to resolve: could not find `PubDiscriminants` in `private`
   --> strum_tests/tests/enum_discriminants.rs:237:18
    |
237 |         private::PubDiscriminants::VariantB,
    |                  ^^^^^^^^^^^^^^^^ could not find `PubDiscriminants` in `private`

error: aborting due to 3 previous errors

For more information about this error, try `rustc --explain E0433`.
error: could not compile `strum_tests`.
```

Peternator7: All right then. Thanks for trying it; let's just leave it as it was :)

linclelinkpart5: Sounds good to me, I'll mark this review as resolved.

Peternator7 (approved):

Peternator7:
Merged; thanks for the PR! I'll work on getting a release out this week.

linclelinkpart5:
Awesome, thanks a bunch!

Peternator7:
Strum `0.20.0` is published. https://crates.io/crates/strum

PR #145, opened 2021-01-15
Replaced ::std with ::core where applicable for no_std projects

rudib:
This makes these procedural macros usable with `no_std` projects.


Peternator7 (approved):

Peternator7:
Looks great! Thanks for the contribution. I'll get a new release out in the next week or so

Issue #147, opened 2021-02-07
Heck Version Upgrade broke 1.31.1


Peternator7:
https://github.com/withoutboats/heck/issues/25 upgraded the msrv to 1.32.0. 

It's probably worth just vendoring/imitating serde's case impl. https://github.com/serde-rs/serde/blob/master/serde_derive/src/internals/case.rs

PR #148, opened 2021-02-08
Use minor version in the README docs.rs links

### LINKED ISSUE #146: Outdated documentation links in README
The links to macros in the docs.rs documentation is outdated because the versions are linked manually.

Since "0.20" is used in the `Cargo.toml` snippet, it may be worth considering using "0.20" in the documentation links as well; docs.rs can resolve crate names just like `cargo` does. This will ensure that the links always point to the intended documentation version. You could also just use "*" if you don't care about keeping the docs.rs links in lockstep with the `Cargo.toml` snippet, as "*" will always point to the latest available version.

I'd be happy to submit a PR with this little change if this issue is assigned to me.

Spaceface16518:
This fix closes #146.

In #146, I mentioned that an asterisk could be used to just link to the latest version. If this is preferred, please request those changes in a review (or like drop a thumbs down on this message or something).

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,14 +68,14 @@ information through strings.
 Strumming is also a very whimsical motion, much like writing Rust code.
 
 [Macro-Renames]: https://github.com/Peternator7/strum/wiki/Macro-Renames
-[EnumString]: https://docs.rs/strum_macros/0.20.0/strum_macros/derive.EnumString.html
-[Display]: https://docs.rs/strum_macros/0.20.0/strum_macros/derive.Display.html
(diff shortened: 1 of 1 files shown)


Spaceface16518:
Not sure why this is failing on rustc 1.31.1, but I'm 99.9% sure it's not related to these changes, so I'm not going to take any action regarding the CI failures.

Peternator7:
Looks great! Thanks for the fix, and yes, the builds are broken because of #147 which I haven't had a chance to fix yet.

PR #149, opened 2021-02-20
Remove the "rename" feature

Peternator7:


dzmitry-lahoda:
what should i use?

Peternator7:
In Rust 2018, you can refer to macros by their path in the crate so you should use the full path to the macro.

```rust
#[derive(strum_macros::EnumString)]
pub YourEnum {}
```

PR #153, opened 2021-03-03
Added a pass-through attribute support to EnumDiscriminants

frol:
The use-cases:

* I wanted to combine `EnumMessage` with `EnumDiscriminants` and found that not all the attributes are passed
* `#[serde]` attributes on discriminant variants could also be useful


Peternator7:
This makes tons of sense to me! Thanks for the PR :) Could you add a few tests in [this file](https://github.com/Peternator7/strum/blob/master/strum_tests/tests/enum_discriminants.rs)

Peternator7:
Looks great!

PR #161, opened 2021-05-14
Replace enum_name with `Self`

enaut:
As it is possible to use constant `Self` instead of the structname
it is less complicated to use `Self`. Using the struct name even
generates warnings on strict warnin-settings.

### DIFF
--- a/strum_macros/src/macros/strings/from_string.rs
+++ b/strum_macros/src/macros/strings/from_string.rs
@@ -87,6 +87,7 @@ pub fn from_string_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
     arms.push(default);
 
     Ok(quote! {
+        #[allow(clippy::use_self)]
         impl #impl_generics ::std::str::FromStr for #name #ty_generics #where_clause {
             type Err = ::strum::ParseError;
             fn from_str(s: &str) -> ::std::result::Result< #name #ty_generics , Self::Err> {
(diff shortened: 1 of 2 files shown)


enaut:
The warning triggered: https://rust-lang.github.io/rust-clippy/master/#use_self

I don't think that the failing CI-builds are due to the changes I made...

Peternator7:
I just fixed the build with #162 . Can you merge from master and re-push? 

I'm not sure `Self` aliases are available in 1.32, but if the build passes then I'm happy to take this change

enaut:
Indeed - as you suspected 1.32 does not support the `Self`. I see that serde adds a lot of `#[automatically_derived]` things to the automatic impls... maybe that would suppress the warning too?

enaut:
Of course the impls could also suppress the warning using: `#[allow(clippy::use_self)]`

what would you prefer?

Peternator7:
I wasn't aware of `#[automatically_derived]`. If that removes the warnings, I'd suggest we do that

enaut:
Unfortunately it does not suppress the warning - I created a minimal project now to test the changes... It was a bit cumbersome with the big project I encountered the problem with. So the only option I see is to add #[allow(clippy::use_self)]. I'll update the pullrequest and see if this compiles.

Peternator7:
No problem! Thanks for making updates as needed. Proc-macros are still a bit mysterious to me as a well :) Merged this PR with the clippy lint, and I'll work on getting a release out in the near future.

PR #164, opened 2021-06-01
fix for field not method complaint in strum_macro's metadata.rs

hlmtre:
[see](https://i.imgur.com/1k131PE.png)

### DIFF
--- a/strum_macros/src/helpers/metadata.rs
+++ b/strum_macros/src/helpers/metadata.rs
@@ -194,7 +194,7 @@ impl Parse for VariantMeta {
             let kw = input.parse()?;
             let value = if input.peek(Token![=]) {
                 let _: Token![=] = input.parse()?;
-                input.parse::<LitBool>()?.value()
+                input.parse::<LitBool>()?.value
             } else {
                 true
             };


Peternator7:
Interesting, do you know what `syn` version you're using? `value()` was added to syn around `1.0.61` [Docs](https://docs.rs/syn/1.0.61/syn/struct.LitBool.html) [Commit](https://github.com/dtolnay/syn/commit/b2f1da284d666ec59571033807f46a8c082330cb#diff-ce4e3f87da05491a74818b9736b02d9c40274befb1e2c13f44c6995eb001b925)

I'm fine accepting this since it appears it will improve compatibility, but there might be an opportunity for another plugin to use a newer version of syn here as well.

hlmtre:
[cargo tree](https://i.imgur.com/AF0U9aw.png) and [inverted, selecting for `syn`](https://i.imgur.com/xnWcr74.png). it appears it's being pulled 'first' for `serde_derive`, which, even on the most recent `master` commit 20 days ago, depends tantalizingly closely on 1.0.60.

i ain't gonna be the guy that suggests a project as big and scary as `serde` update their `syn` dependency :cold_sweat:

Peternator7:
How disappointing! This is an easy enough workaround though and lower stakes than updating serde-derive as you said, so I'll go ahead and merge this change in.

hlmtre:
Thank you!

PR #167, opened 2021-06-04
Use proper minimum syn version

e00E:
RE: https://github.com/Peternator7/strum/pull/164

The correct way to fix this is to specify that you need this minimum syn version in this crate. https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html . In strum-macro's Cargo.toml change it to 

```toml
syn = { version = "1.0.61", features = ["parsing", "extra-traits"] }
```

serde_derive is not actually forcing an old version. It accepts any version >= 1.0 https://github.com/serde-rs/serde/blob/985725f820a08fbe1c23688422d79200d24502ec/serde_derive/Cargo.toml#L28 .

What probably happened is that in @hlmtre 's project the Cargo.lock file contained the older version and since strum doesn't spewcify which version is needed Cargo kept the older one.. To fix this locally before this proper fix is in you could have run `cargo update`.

### DIFF
--- a/strum_macros/Cargo.toml
+++ b/strum_macros/Cargo.toml
@@ -22,7 +22,7 @@ name = "strum_macros"
 heck = "0.3"
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = { version = "1.0", features = ["parsing", "extra-traits"] }
+syn = { version = "1.0.61", features = ["parsing", "extra-traits"] }
 
 [dev-dependencies]
 strum = "0.20"


hlmtre:
@e00E thank you for the correction. I think that's exactly what happened, and I didn't know enough about the cargo dependency system to know how to pull a newer version that still satisfied the dependency.

Peternator7:
Hi @e00E,

Thanks for the explanation! That does make sense.

Given @hlmtre's change improved backwards compatibility, I'm not sure it's necessary now to pin the minimum version. What do you think the benefit now is to adding an additional constraint?

e00E:
The minimum version should be whatever minimum version of syn you need based on what syn features you are using. If you know that the project would still compile with syn 1.0.0 then you can keep the current version requirement.
On the other hand there is really nothing wrong with requiring a newer version. The beauty of semantic versioning is that newer versions are backward compatible. So if you specify a more recent version of syn then you will never break people's projects. The only downside is potential minimum rust compiler version changes but I don't think that is relevant here.
Also cargo will usually prefer newer compatible versions anyway because they can contain bug fixes.

I feel it is going to be easier to use the newest version of syn instead of manually keeping track of what the true minimum version based on what syn features are used in the code. But feel free to close the PR if you have a different preference. I mostly opened it to explain what was going on.

Peternator7:
@e00E , I very much appreciate the explanation, but I am going to continue tracking minimum versions as much as possible. Since it's mostly a convenience library, I try to target maximum compatibility with older rust versions and packages to avoid breaking people

PR #170, opened 2021-06-30
Re-export macros

billy1624:
## Motivation

Our team is developing a library which requires downstream end-user to derive `strum_macro::EnumIter` for their custom enum.

```rust
pub trait RelationTrait: Iterable + Debug + 'static { ... }

// Derive EnumIter
#[derive(Copy, Clone, Debug, EnumIter)]
pub enum Relation {
    Cake,
    Filling,
}
```

However, the expanded codes from `strum_macro::EnumIter` looks something like this and failed to compile...
This signature, `::strum::IntoEnumIterator`, implies our downstream end-user must add the `strum` dependency on their own.

```rust
impl ::strum::IntoEnumIterator for Relation { ... }
```

```
error[E0433]: failed to resolve: could not find `strum` in the list of imported crates
   |
39 | #[derive(Copy, Clone, Debug, EnumIter)]
   |                              ^^^^^^^^ could not find `strum` in the list of imported crates
   |
```

Hence, we want to find a way to re-export `strum` & `strum_macros` in our library. And downstream end-user don't need to specify the `strum` dependency on their own.

## Proposed Solution

Adding an optional derive attribute `#[strum(crate_path = "some_crate::strum")]`, for example...

```rust
// Derive EnumIter with crate_path derive attribute
#[derive(Copy, Clone, Debug, EnumIter)]
#[strum(crate_path = "some_crate::strum")]
pub enum Relation {
    Cake,
    Filling,
}

// Derive macro will be expanded into
impl some_crate::strum::IntoEnumIterator for Relation { ... }
```

Backward compatibility, if the derive attribute was not provided...

```rust
// Derive EnumIter without crate_path derive attribute
#[derive(Copy, Clone, Debug, EnumIter)]
pub enum Relation {
    Cake,
    Filling,
}

// Derive macro will be expanded into
impl ::strum::IntoEnumIterator for Relation { ... }
```

## References

Similar feature was discussed on serde-rs/serde#1465. And `serde` provides a optional derive attribute to achieve it.

```rust
#[derive(Copy, Clone, Debug, Deserialize, Serialize)]
#[serde(crate = "some_crate::serde")]
struct Vertex {
    position: [u32; 3],
}

```

### DIFF
--- a/strum_macros/src/helpers/metadata.rs
+++ b/strum_macros/src/helpers/metadata.rs
@@ -2,6 +2,7 @@ use proc_macro2::{Span, TokenStream};
 use syn::{
     parenthesized,
     parse::{Parse, ParseStream},
+    parse2, parse_str,
     punctuated::Punctuated,
     spanned::Spanned,
     Attribute, DeriveInput, Ident, LitBool, LitStr, Path, Token, Variant, Visibility,
@@ -14,6 +15,7 @@ pub mod kw {
 
     // enum metadata
     custom_keyword!(serialize_all);
+    custom_keyword!(crate_path);
 
     // enum discriminant metadata
     custom_keyword!(derive);
@@ -37,6 +39,10 @@ pub enum EnumMeta {
         case_style: CaseStyle,
     },
     AsciiCaseInsensitive(kw::ascii_case_insensitive),
+    CratePath {
+        kw: kw::crate_path,
+        path: Path,
+    },
 }
 
 impl Parse for EnumMeta {
@@ -47,6 +53,13 @@ impl Parse for EnumMeta {
             input.parse::<Token![=]>()?;
             let case_style = input.parse()?;
             Ok(EnumMeta::SerializeAll { kw, case_style })
+        } else if lookahead.peek(kw::crate_path) {
+            let kw = input.parse::<kw::crate_path>()?;
+            input.parse::<Token![=]>()?;
+            let path_str: LitStr = input.parse()?;
+            let path_tokens = parse_str(&path_str.value())?;
+            let path = parse2(path_tokens)?;
+            Ok(EnumMeta::CratePath { kw, path })
         } else if lookahead.peek(kw::ascii_case_insensitive) {
             let kw = input.parse()?;
             Ok(EnumMeta::AsciiCaseInsensitive(kw))
@@ -61,6 +74,7 @@ impl Spanned for EnumMeta {
         match self {
             EnumMeta::SerializeAll { kw, .. } => kw.span(),
             EnumMeta::AsciiCaseInsensitive(kw) => kw.span(),
+            EnumMeta::CratePath { kw, .. } => kw.span(),
         }
     }
 }
--- a/strum_macros/src/helpers/metadata.rs
+++ b/strum_macros/src/helpers/metadata.rs
@@ -15,7 +15,7 @@ pub mod kw {
 
     // enum metadata
     custom_keyword!(serialize_all);
-    custom_keyword!(crate_path);
+    custom_keyword!(Crate);
 
     // enum discriminant metadata
     custom_keyword!(derive);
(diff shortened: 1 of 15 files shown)


strum_macros/src/helpers/metadata.rs
```diff
@@ -14,6 +15,7 @@ pub mod kw {
 
     // enum metadata
     custom_keyword!(serialize_all);
+    custom_keyword!(crate_path);
```
Peternator7, line 18: Did you say serde uses "crate?" Can we also do that for consistency?

strum_macros/src/macros/enum_count.rs
```diff
@@ -2,13 +2,15 @@ use proc_macro2::TokenStream;
...
-use crate::helpers::non_enum_error;
+use crate::helpers::{non_enum_error, HasTypeProperties};
 
 pub(crate) fn enum_count_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
     let n = match &ast.data {
         Data::Enum(v) => v.variants.len(),
         _ => return Err(non_enum_error()),
     };
+    let type_properties = ast.get_type_properties()?;
+    let strum = type_properties.get_crate_path();
```
Peternator7, line 13: Can you find a more descriptive name here than strum. Maybe `strum_module_path` or something

strum_macros/src/macros/enum_iter.rs
```diff
@@ -2,13 +2,15 @@ use proc_macro2::{Span, TokenStream};
...
-use crate::helpers::{non_enum_error, HasStrumVariantProperties};
+use crate::helpers::{non_enum_error, HasStrumVariantProperties, HasTypeProperties};
 
 pub fn enum_iter_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
     let name = &ast.ident;
     let gen = &ast.generics;
     let (impl_generics, ty_generics, where_clause) = gen.split_for_impl();
     let vis = &ast.vis;
+    let type_properties = ast.get_type_properties()?;
+    let strum = type_properties.get_crate_path();
```
Peternator7, line 13: Same comment as above

Peternator7 (changes requested):
This makes a lot of sense to me! Thanks for putting together a PR :) I left a few comments and if you could add a few tests, that would be great.

Maybe just try re-exporting the strum derives from some nested module in the tests and make sure it still works. You could also use the [debugging](https://github.com/Peternator7/strum#debugging) section to make sure the proper code is generated.

strum_macros/src/helpers/metadata.rs
billy1624: `crate` is a Rust keyword, is it okay if I use `Crate` instead?

strum_macros/src/macros/enum_count.rs
billy1624: Changed to
```rust
let strum_module_path = type_properties.crate_module_path();
```

billy1624:
Please review @Peternator7

strum_macros/src/helpers/metadata.rs
Peternator7: sorry for the delay getting back to you. thanks for your patience. I would prefer crate if we can since that's what serde does. Is there a technical reason we can't do that?

billy1624: Just overcome the technical difficulties, see the latest commit, now we have

```rust
#[derive(Debug, Eq, PartialEq, EnumIter)]
#[strum(crate = "nested::module::strum")]
enum Week {
    Sunday,
    Monday,
    Tuesday,
    Wednesday,
    Thursday,
    Friday,
    Saturday,
}
```

Peternator7 (approved):

Peternator7:
lgtm! Thanks for the contribution. I'm on vacation next week, but I'll try and get a release out if I find a little time!

billy1624:
Have a nice vacation!!

Peternator7:
Hi @cosmicchipsocket and @billy1624, thanks for the push. Version `0.22` is live on crates.io

billy1624:
Thanks! @cosmicchipsocket

PR #171, opened 2021-07-30
Features/no std

luctius:
Adds a 'std' default feature, to allow no_std use.


luctius:
I am unsure what to do about the CI errors. Any advice?

Peternator7:
I'm flipping through some issues and it looks like Cargo might've updated it's `curl` version. Let me see if a newer appveyor base image fixes the issue. If not, I'll open an issue against cargo.

Peternator7:
Fixed the build issues in another PR by updating to a more recent Visual Studio toolkit. This looks good to me 👍

PR #174, opened 2021-08-16
Fix for Syn version spec ("1.0" → "1.0.61")

Tamschi:
The dependency build of `strum_macros` [fails](https://github.com/Tamschi/reserde/pull/8/checks?check_run_id=3342430840) in my project after `cargo +nightly update -Z minimal-versions`:

```
   Compiling syn v1.0.60
   […]
   Compiling strum_macros v0.21.0
error[E0599]: no method named `value` found for struct `LitBool` in the current scope
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/strum_macros-0.21.0/src/helpers/metadata.rs:197:43
    |
197 |                 input.parse::<LitBool>()?.value()
    |                                           ^^^^^-- help: remove the arguments
    |                                           |
    |                                           field, not a method

For more information about this error, try `rustc --explain E0599`.
error: could not compile `strum_macros` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```

This is due to the `syn = { version = "1.0", features = ["parsing", "extra-traits"] }` dependency [in `strum_macros/Cargo.toml`](https://github.com/Peternator7/strum/blob/master/strum_macros/Cargo.toml#L25), which isn't quite specific enough in this case as Syn doesn't follow Semver and regularly has features added in patch versions.

`LitBool.value()` was added in `syn@1.0.61`: `https://github.com/dtolnay/syn/compare/1.0.60...1.0.61#diff-ce4e3f87da05491a74818b9736b02d9c40274befb1e2c13f44c6995eb001b925R513-R515`
(Apologies, GitHub refuses to parse that link correctly.)

I've confirmed locally that version 1.0.61 is enough to fix my build.

### DIFF
--- a/strum_macros/Cargo.toml
+++ b/strum_macros/Cargo.toml
@@ -22,7 +22,7 @@ name = "strum_macros"
 heck = "0.3"
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = { version = "1.0", features = ["parsing", "extra-traits"] }
+syn = { version = "1.0.61", features = ["parsing", "extra-traits"] }
 
 [dev-dependencies]
 strum = "0.20"


Peternator7:
Hi @Tamschi, thanks for this PR. I know #164 fixed a similiar issue, and that was released in strum_macros `1.21.1`. Do you happen to know what version you were using?

If this a different occurrence, I would prefer we fix it in the same was as #164 because it strictly improves compatibility.

Tamschi:
Ah, I didn't realise that was a public field too from the changelog, sorry about that.
I was using `strum_macros` at `0.21.0`, yes, but as indirect dependency of course, so updating that would still be a workaround on my end. 

I just verified that

```toml
[dependencies]
strum = { version = "0.21.0", features = ["derive"], default-features = false }

[build-dependencies]
strum_macros = "0.21.1"
```

resolves `syn` to `1.0.0` with minimal versions and builds fine that way. (So I'll close this PR since no actual code change is necessary.)

If you could yank `strum_macros`'s `0.21.0`, that would fix the underlying issue for everyone who newly resolves their lockfile.
There's no need to update the `strum` crate and any projects with already locked builds should be unaffected by this too.

Tamschi:
I'd normally update the PR to add the check for this to Travis or AppVeyor, but I'm not familiar enough with those.
It might be a way to fix the CI for Rust `1.32.0` though, since you probably have a dependency with a looser minimal Rust version policy.

PR #178, opened 2021-09-24
Fixed incorrectly commented `deprecated` attribute

SOF3:


Peternator7:
Welp that's unfortunate, since this is just the trait and not the macro itself, could you update the the "since" to 0.22. That'll be the next version and this error would likely confuse someone if it's back-dated.

SOF3:
Hmm? I didn't actually change any behaviour, I just fixed a possible typo. But sure, I would just copy over.

strum/src/lib.rs
```diff
@@ -168,7 +168,7 @@ pub trait EnumProperty {
 
 /// A cheap reference-to-reference conversion. Used to convert a value to a
 /// reference value with `'static` lifetime within generic code.
-/// #[deprecated(since="0.13.0", note="please use `#[derive(IntoStaticStr)]` instead")]
+#[deprecated(since="0.13.0", note="please use `#[derive(IntoStaticStr)]` instead")]
```
Peternator7, line 171: Sorry, I may not have been clear. Can you make this `since = "0.22.0"` instead of "0.13"

strum_macros/src/lib.rs
```diff
@@ -191,6 +191,7 @@ pub fn variant_names(input: proc_macro::TokenStream) -> proc_macro::TokenStream
 }
 
 #[proc_macro_derive(AsStaticStr, attributes(strum))]
+#[deprecated(since="0.13.0", note="please use `#[derive(IntoStaticStr)]` instead")]
```
Peternator7, line 194: This one as well

Peternator7 (commented):
Thanks for the update; sorry if I had been confusing. I had meant that we should update the "deprecation" notice to `0.22` instead of `0.13` because it wasn't doing anything until now.

SOF3:
Resolved.

Peternator7:
Thanks!

PR #185, opened 2021-10-16
EnumIndex implementation

drewkett:
@Peternator7 This is an initial pass at an implementation of an index function that is `const`. The following lists some of the outstanding items. In addition, I'm really not sure I captured all of the desired behavior in that macro implementation. 

- [ ] Is this the best name for the macro name or the method name (currently its `index` and `const_index`)
  - [ ] Should there by `try_` variants of the functions so that the non-`try` versions panic on unknown value similar to `Index` trait returns an `Option`?
- [x] Add documentation
- [x] The argument to the get/index function should be the same type as the `repr` (ie if the enum is `repr(u8)`)
- [ ] Add ability to rename get/index function?
- Discriminant
  - [x] Add error messages related to discriminant code
  - [ ] Make sure the implementation is correct beyond the basic tests that are currently written.


drewkett:
I went ahead and updated some things.
- Added documentation
- Renamed macro to `EnumIndex`
- The macro now generates up to two functions. `index` in all cases (that don't generate an error due to the existence of lifetimes). And `const_index` when all the variants are unit variants (`Default::default` is not `const`). 
- Renamed macro to `EnumIndex`

drewkett:
I got a version working to handle custom `repr` working. The code needs cleaning up in general but just wanted to get something committed that was working, since it ended up being a bit more complicated that I thought it would be

drewkett:
Think I fixed everything mentioned. I added `rustversion` as well. The only weird thing was the I made the `const_index` doc tests conditional on the rust version. Not sure if you have any other ideas for that.

Peternator7:
Combining them is certainly interesting. One obstacle is that `EnumIndex` requires variant data have a `Default` impl where `EnumDiscriminants` doesn't since it's strictly Enum -> Discriminant. That alone might be enough to justify separating them. 

More generally, I wonder if the the 2 macros serve different users:

- People using `EnumDiscriminants` probably have complicated enum types that they want to be able to reduce to just the discriminant. Since rust-lang/rust#60553 is open, they probably aren't using a custom repr or explicit discriminants on these enums. These enums strike me as unconcerned with the representation of their discriminant. It's an implementation detail mostly.

- My guess is that users of `EnumIndex` probably have simple enum types with few or no data fields; they also might have a custom repr and explicit integral discriminant values. In this case, the representation seems like it might be one of the defining features of the particular enum.

I could certainly be wrong and feel free to disagree, but I think they are probably best as 2 different macros because they are interesting in different scenarios.

So having said all that, I'm still in favor of `FromRepr`, but open to other ideas if anything comes to mind.

drewkett:
Looking at more things, I agree with you. I came across this [PR](https://github.com/rust-lang/rust/pull/81642) which is adding a `FromRepr` trait to rust only for repr enums without any data. Since this feature is fundamentally similar it should probably be the same name with similar behavior. The three deviations this library would have compared to the above feature would be: 1) add the ability to use it with any that implements Default, 2) having `from_repr` be able to be `const` 3) its a safe function that returns None on invalid repr. All of those might be useful to people.  I'll go ahead and update the name of the macro/function

Peternator7 (approved):

Peternator7:
Everything looks good! Thanks for all the work on this. I'll try and get a release out in the next week or so with this and #186

drewkett:
It was a pleasure working with you on this. Thanks from a happy user of the library.

Peternator7:
Glad it works well for you! :D 

Just release 0.23 on crates. https://crates.io/crates/strum

PR #186, opened 2021-10-19
Add TryFrom<&str> to EnumString

### LINKED ISSUE #107: Implement `std::convert::TryFrom<&str>` for `EnumString`
The `EnumString` macro provides `std::str::FromStr`. 

I wonder whether it would be possible to also provide an implementation of the nearly equivalent `std::convert::TryFrom<&str>`?

dspicher:
The implementation simply delegates to `std::str::FromStr`.

To maintain compatibility with Rust 1.32, a new dependency
is introduced which allows us to emit the `TryFrom<&str>`
implementation only for Rust 1.34 and above when
`std::convert::TryFrom` was stabilized.

Closes #107.

### DIFF
--- a/strum_macros/Cargo.toml
+++ b/strum_macros/Cargo.toml
@@ -22,6 +22,7 @@ name = "strum_macros"
 heck = "0.3"
 proc-macro2 = "1.0"
 quote = "1.0"
+rustversion = "1.0"
 syn = { version = "1.0", features = ["parsing", "extra-traits"] }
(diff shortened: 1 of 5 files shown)


dspicher:
Thanks for the fast review @Peternator7, I just force-pushed some fixes.

Peternator7:
Released as part of 0.23. Thanks for the PR. https://crates.io/crates/strum

PR #189, opened 2021-10-25
Use core::result instead of std::result in order to make it work in a…

daym:
… no_std environment.

This pull request would avoid using `std::result`, instead favoring `core::result`, in order to make `FromStr` work in a `no_std` environment.


Peternator7:
Released in version 0.23! Thanks for the PR. https://crates.io/crates/strum

daym:
Thank you for the release.

Unfortunately, it does not work since there are now more references to std::convert and so on in the generated code.

I recommend adding a simple integration test in the no_std environment.

Peternator7:
Sorry about that, merged the other PR. Will add integration test when there's time.

PR #192, opened 2021-11-18
Use core::* instead of std::* in order to make it work in a no_std en…

daym:
…vironment.

I previously had fixed it via https://github.com/Peternator7/strum/pull/189 , but meanwhile other uses of `std` inside `quote!` have been merged. Hence fixing it again here.


Peternator7:
Sorry about this. Pushed a patch for [strum_macros](https://crates.io/crates/strum_macros) and added some tests in #196

daym:
Thank you!

PR #193, opened 2021-11-18
Disambiguate associated items in trait impls

tylerwhall:
The addition of `TryFrom<&str>` for `EnumString` breaks for enums that have
a variant called "Error". Rewrite as `<Self as Trait>` where associated
types are used as the return type for trait functions.

Example:

```rust
#[derive(EnumString)]
pub enum Test {
    Error,
}
```

error: ambiguous associated item
   --> src/lib.rs:3:10
    |
3   | #[derive(EnumString)]
    |          ^^^^^^^^^^
    |
    = note: `#[deny(ambiguous_associated_items)]` on by default
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #57644 <https://github.com/rust-lang/rust/issues/57644>

### DIFF
--- a/strum_macros/src/macros/strings/from_string.rs
+++ b/strum_macros/src/macros/strings/from_string.rs
@@ -91,7 +91,7 @@ pub fn from_string_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
         #[allow(clippy::use_self)]
         impl #impl_generics ::core::str::FromStr for #name #ty_generics #where_clause {
             type Err = #strum_module_path::ParseError;
-            fn from_str(s: &str) -> ::core::result::Result< #name #ty_generics , Self::Err> {
+            fn from_str(s: &str) -> ::core::result::Result< #name #ty_generics , <Self as ::core::str::FromStr>::Err> {
                 match s {
                     #(#arms),*
                 }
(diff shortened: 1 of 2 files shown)


jameswald (approved):

strum_macros/src/macros/strings/from_string.rs
```diff
@@ -136,7 +136,7 @@ fn try_from_str(
         #[allow(clippy::use_self)]
         impl #impl_generics ::std::convert::TryFrom<&str> for #name #ty_generics #where_clause {
             type Error = #strum_module_path::ParseError;
-            fn try_from(s: &str) -> ::std::result::Result< #name #ty_generics , Self::Error> {
+            fn try_from(s: &str) -> ::std::result::Result< #name #ty_generics , <Self as ::std::convert::TryFrom<&str>>::Error> {
```
Peternator7, line 139: Can you switch this to `::core::convert`? #189 was supposed to fix this and then #192 fixed it again.

Peternator7:
Makes sense. If you can resolve the conflict, I can get this merged and put out a patch release.

strum_macros/src/macros/strings/from_string.rs
tylerwhall: Fixed. Thanks for supporting no_std.

Peternator7:
Just published a patch version of [strum_macros](https://crates.io/crates/strum_macros) to crates.io

PR #201, opened 2022-01-04
Update to heck 0.4

alex:


alex:
It's the 1.32 builds that are failing, but I don't think they're related to this PR?

Peternator7:
Looks good! Agree that the error looks unrelated. Let me fix the build and then we can get this merged :)

Peternator7:
Tests pass locally after fixing the dependency version. Thanks for the update

alex:
Thanks!

PR #202, opened 2022-01-07
Add doc comment to FromRepr

ljoonal:
Currently when deriving FromRepr for an enum with `#![deny(missing_docs)]` will cause errors.
This change would simple add a doc comment to the generated from_repr function to avoid the error.
An alternative I considered would be to just add `#[allow(missing_docs)]`.


Peternator7:
LGTM 👍 Thanks for the fix!

PR #204, opened 2022-01-08
Peternator7/enum iter doc comment

### LINKED ISSUE #200: EnumIter macro causes an error with #[forbid(missing_docs)]
When attempting to use the `EnumIter` derive macro in a crate with #[forbid(missing_docs)], I receive the following error:

> allow(missing_docs) incompatible with previous forbid
> overruled by previous forbid

The error seems to be somehow caused by the generated code, but I don't see any place where you are explicitly allowing it 🤔

Peternator7:
Closes #200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.23.1
+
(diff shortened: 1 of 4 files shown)


alice-i-cecile:
Thanks!

PR #207, opened 2022-01-18
an initial prototype of an enum  metadata trait

ratmice:
This is kind of a trait/associated type pretzel, but before you run in horror please check out the tests...
They at least show how it is used and that the whole horror show underneath it doesn't really impact the usage...

There are still some FIXME's in here,

There are quite a few additional constraints than e.g. FromRepr:
* discriminants are comprised of unique bits
* An assumption in the iterator size hint that discriminants have a single 1 bit.
* no zero discriminant
* ~~The enum type is clone.~~  Edit: Fixed this

Edit: In theory at least we could relax uniqueness to 'has at least 1 unique bit' maybe, but this requires a bit fancier macro implementation.

There are some FIXME's where these constraints aren't actually enforced.

Given that there are quite a few macro packages that do similar things, totally understand if it isn't something you want
to bring into your package, but thought I would offer it (Since I don't really know proc-macros, and just cribbed the macro side from `from_repr.rs`,

If you would rather close it, I can just throw it into my own for my own usage...
Thanks

FWIW this is my first stab at generalizing it from my existing code base, because I needed to split it up one enum into two.
In generalizing it is quite possible simplifications that can be made regarding the trait typing, but for now I'm just happy it typechecks and runs.


ratmice:
Oops, I think there is a case where this can return an infinite iterator, when the enum saturates the repr, and the mask is entirely ones...

But that should be fixable by adding a `const REPR_SIZE` to the trait and setting that in the macro.

ratmice:
So assuming tests pass, which I believe they should I think this might suffice as a proof of concept,
Though I think that the iterator could be better, that would require a bunch of changes to the trait requirements,

So it might not be worth going through trait for trait, on the every trait bound but the overall idea.

ratmice:
So, I think i'm done for now, so marking ready for review.

Things like macros for marker traits can really wait, because they only allow us to optimize the existing iterator.
Now that i've redone the size_hint not to make those assumptions.

ratmice:
So I'm thinking that the rest of the arithmetic ops perhaps might not be useful as written, because they take an `OpaqueRepr` as RHS rather than e.g. a `Self::R`, but haven't really investigated or thought through it much.

Peternator7:
Hi @ratmice, I think the `OpaqueRepr` work is very clever, but I'm not sure it's the best fit for this project. I don't want to take a dependency on `num-traits`. There's also lots of different opinions about the best way to handle bitflags, and I like your design, but I don't think it's something I want `strum` to have an opinion about. I would suggest you refactor it into its own crate.

The rest of `EnumMetadata` looks pretty good though if you're still interested in driving this PR. I'd like if we can share code from the existing impl's rather than duplicating it, but otherwise, that looks pretty good to me. Thanks!

ratmice:
Fair enough!

I'll have a look at it, I think getting rid getting rid of `num_traits`, and seeing how the `Repr` type fairs without it, and if there is any fallout for the base trait impl, I don't see a problem offhand keeping the `Repr` type, and dropping num_traits if possible.

Which would make it possible to keep the generic trait based to/from impl.

Further I don't know exactly whether it will be possible to share code, Really this is my first project using macros so i'm not exactly in my comfort zone there. 

I think it will be easier and a lot cleaner to do on a new branch/pr, so I'm going to close and start a new one!

ratmice:
One last thing I think to note here, we could in theory keep `OpaqueRepr` without using `num_traits`, I don't think that my `MaskIterator` used any of num_traits any longer.

But I fully undestand if you don't want any of the `impl ... for OpaqueRepr` which may impact compilation times even if it is just depending upon core.

Issue #183, opened 2021-10-13
Proposal: Condense several "Derives" together


Peternator7:
`EnumCount`, `EnumVariants` and `EnumString` are all always implementable (as opposed to something like `EnumIter` which will fail if a variant can't be constructed with `Default::default`).

Should we bundle them into one derive + trait. The implementations are pretty small so it isn't much overhead, and it makes the docs a little easier. Implement this one thing to get all these features.

```rust
pub trait EnumMetadata {
    const VARIANT_COUNT: usize;
    const VARIANTS: &'static &'static str;

    fn variant_name(&self) -> &'static str; 
}
```

PR #208, opened 2022-01-24
minimal extraction of enum_metadata from PR #207.

ratmice:
This is the minimal extraction of `EnumMetadata`, which excises `OpaqueRepr` from my pr #207.

It doesn't yet fix the duplicate code issue, I also think there is a function in #183 which this hadn't implemented.

I need to work on the other crate where the stuff which was excised goes however,
So If anyone wants to pick this up, by all means.

I assume what needs to be done is pull as much as we can out into `strum_macros/helpers/` and then call those from the various `..._inner(..)` calls.  Instead of just duplicating everything.  But note that the types and constness between the inherent and trait impl may differ making it difficult to remove all duplication.

Otherwise i'll try and get back to this when I can -- or we can close it and reopen at that time!

### DIFF
--- a/strum/src/lib.rs
+++ b/strum/src/lib.rs
@@ -30,6 +30,8 @@
 // only for documentation purposes
 pub mod additional_attributes;
 
+use core::ops;
+
 /// The ParseError enum is a collection of all the possible reasons
 /// an enum can fail to parse from a string.
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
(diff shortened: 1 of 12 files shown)


ratmice:
It seems like github's diff comparison doesn't pick up on the fact that macros/enum_metadata.rs is a copy of enum_fromrepr.rs with 70% similarity, but viewing in another diff tool like gitk should be helpful when getting rid of duplicated code.

ratmice:
So, I took a first pass at this, somehow my attempt to move out the duplicated code ended up longer than the combined duplication.
It might not be very pretty, likely requires another pass

ratmice:
Wonder if we should include a `const  REPR_TYPE_STR: &'static str = "u8"` for example.

schneems:
The original referenced PR was closed and this one was last updated in 2022. What is the status? Is this still relevant or needed? 

> Otherwise i'll try and get back to this when I can -- or we can close it and reopen at that time!

❤️  If this isn't actively being worked on can it be closed? If you want to come back to it, I think it would benefit from a brand new PR with a self contained description explaining either what this is fixing (expected/actual disagreement) or what it's adding (can't do today versus with my feature you can).

Can you close this PR?

ratmice:
Sure this can be closed, It works fine, but never got reviewed.
This patch is used in my library https://github.com/ratmice/enum_extra

but I'll likely just rewrite that to avoid using strum, from the comment in my original PR it seemed like the author didn't want the additional complexity, and I never heard back after trying to minimize that in this patch...

Until I rewrite that crate to not use this branch, I'll just use my local fork of strum.

ratmice:
To answer your question about what this does that strum currently can't do today,
this crate basically provided what is now https://docs.rs/strum/latest/strum/trait.VariantMetadata.html

beyond what is in that it also added an associated `type Repr` that gave you a type listed in the `#[repr(..)]` where the enum was declared, `from_repr` and `to_repr` functions that converted between the repr type and the enum type.

PR #214, opened 2022-03-14
Use full path for `None` in `EnumIter`

spenserblack:
It was possible for a conflict with other `None` identities.

### Minimal Reproduction

```toml
[dependencies]
strum = { version = "0.24.0", features = ["derive"] }
```

```rust
use strum::EnumIter;
use Foo::*;

#[derive(EnumIter)]
enum Foo {
    None,
}
```


spenserblack:
:thinking: Tests are failing, but `git bisect` points to 94460f006cc6ecce6fe77262a073372d283ad0fe

PR #217, opened 2022-04-01
Add TryFrom to FromRepr

danog:
Similar to #186, automatically implements TryFrom based on the discriminant type.


Peternator7:
Hello @danog, Thanks for the PR! I'm onboard with the general idea. The test failures in 1.32 look legitimate so if you can fix those, we can merge this.

Peternator7:
Hello @danog, Thanks for the PR! I'm onboard with the general idea. The test failures in 1.32 look legitimate so if you can fix those, we can merge this.

danog:
This should do the trick! :D

PR #228, opened 2022-06-19
Revert "Add TryFrom to FromRepr (#217)"

Peternator7:
This reverts commit ac757fa9702cacd4559c444b309c27b97fa1de59.

Adding TryFrom to `FromRepr` and it was more problematic than I had originally thought for people who hit this issue to work around it. Reverting for now so that we can rethink how it should be implemented.


Peternator7:
@danog for fyi, adding a new trait impl was a breaking change that didn't have an especially easy mitigation so I'm rolling this back until we have a better solution - either a separate plugin or flag that enables/disables the feature.

PR #250, opened 2023-01-07
Fix camel_case to PascalCase bug

bouzuya:
Specifying camel_case for serialize_all will result in a PascalCase. Is this the intended behavior?

This pull request assumes that this is unintended behavior and fixes it.

---

This is not related to this pull request, but I have a point of concern. If you allow a snake_case specification for each variant, I would like to see it supported for all variants. Otherwise, I think it would be a good idea to reject the `camel_case` and `kebab_case` specifications.

https://github.com/Peternator7/strum/pull/250/files#diff-801759161311a5149bd0d6c227fea4237bb6d8a8bb9e84f4703d305e2f131ae1R129

The serde crate rename_all attribute appears to reject it.

https://github.com/serde-rs/serde/blob/ce0844b9ecc32377b5e4545d759d385a8c46bc6a/serde_derive/src/internals/case.rs#L38-L59

### DIFF
--- a/strum_macros/src/helpers/case_style.rs
+++ b/strum_macros/src/helpers/case_style.rs
@@ -57,12 +57,14 @@ impl FromStr for CaseStyle {
 
     fn from_str(text: &str) -> Result<Self, ()> {
         Ok(match text {
-            "camel_case" | "PascalCase" => CaseStyle::PascalCase,
+            // "camel_case" is a soft-deprecated case-style left for backward compatibility.
+            // <https://github.com/Peternator7/strum/pull/250#issuecomment-1374682221>
+            "PascalCase" | "camel_case" => CaseStyle::PascalCase,
             "camelCase" => CaseStyle::CamelCase,
             "snake_case" | "snek_case" => CaseStyle::SnakeCase,
-            "kebab_case" | "kebab-case" => CaseStyle::KebabCase,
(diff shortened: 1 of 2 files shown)


Peternator7:
Hi @bouzuya, I think this is an awkward case where it's by design. Heck (which provides the casing) refers to the cases as lowerCamelCase and UpperCamelCase so the original names for the case styles followed it's convention. As you noticed, the newer case-styles follow the pattern serde does of using the style itself in the name and that follows the more generally used convention of camelCase and PascalCase.

[The docs](https://docs.rs/strum/latest/strum/additional_attributes/index.html#attributes-on-enums) recommend using the new naming format so I consider the old names to be soft-deprecated, but I don't think it's worth causing a breaking change by removing the old names or what they do.

bouzuya:
Thank you for your answer. I understood as follows.

- Some people interpret `camel_case` as `(upper_)camel_case` (=`PascalCase`)
- Names like `camel_case` are soft-deprecated
- The documentation recommends `camelCase`.
- Breaking changes by removing `camel_case` is not good for backward compatibility; removing `camel_case` is not worth it.

Based on your answer, I am going to stop changing the behavior (removing `camel_case`) and leave a comment stating that it is soft-deprecated.

Peternator7 (approved):

PR #253, opened 2023-02-26
Fix doc comment for strum::EnumIter macro

kesyog:
The `EnumIter` macro's doc comment has a typo around the implemented trait. The implemented trait is `strum::IntoEnumIterator`.


Peternator7:
Good catch. Thanks!

Issue #245, opened 2022-10-16
Feature Request: Add attribute for types that don't implement Default


ewoolsey:
Heres a quick example of how it could look:
```
#[derive(EnumIter)]
pub enum AssetInfo {
    #[strum(default=Addr::unchecked(""))]
    Token { contract_addr: Addr },
    NativeToken { denom: String },
}
```
This would be very useful when working with types from external crates, or when implementing Default doesn't otherwise make sense.

Peternator7:
Hey @ewoolsey, there have been a few similiar requests, and I think this would be a very good fit for `strum`, but I haven't had time to implement it yet. If you're interested, I'd be happy to accept a PR, otherwise, I'll work on it when I have time.

Peternator7:
I think I would err towards a design like the one serde uses `#[serde(serialize_with = "path")]` so that it feels familiar to people. Probably `#[strum(default_with = "path")] where path is the path to a function that accepts 0 arguments and returns a value of type T

ewoolsey:
@Peternator7 since posting this I’ve found an alternative solution (for my specific use case) with the clap Subcommand derive macro. My workload is pretty high the next few weeks so I’m not sure I’ll have time to contribute for a while, although I would like to get my hands dirty with some more involved proc macro projects. If you decide to pickup the torch on this let me know, I do think this would be a great addition to strum.

ericmcbride:
I will knock this out

ericmcbride:
I got a working proto type of this at the Variant Level + The Inner Variant Fields. Tests all pass and I didn't break anything.   I plan on refactoring a little bit of the code, so Its easier to add Non Derive / Variant level macros, if its required in the future.

PR #254, opened 2023-02-26
Add default_with strum macro

ericmcbride:
- My first shot at messing with Syn + Quote + Proc Macros in general
- Added the default_with macro, from #245 
- Added tests to cover new macro
- Will Add doc strings after verifying if my approach is ok

### DIFF
--- a/strum_macros/src/macros/enum_messages.rs
+++ b/strum_macros/src/macros/enum_messages.rs
@@ -74,14 +74,17 @@ pub fn enum_message_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
         if !documentation.is_empty() {
             let params = params.clone();
             // Strip a single leading space from each documentation line.
-            let documentation: Vec<LitStr> = documentation.iter().map(|lit_str| {
-                let line = lit_str.value();
-                if line.starts_with(' ') {
-                    LitStr::new(&line.as_str()[1..], lit_str.span())
-                } else {
-                    lit_str.clone()
-                }
-            }).collect();
+            let documentation: Vec<LitStr> = documentation
+                .iter()
+                .map(|lit_str| {
+                    let line = lit_str.value();
+                    if line.starts_with(' ') {
+                        LitStr::new(&line.as_str()[1..], lit_str.span())
+                    } else {
+                        lit_str.clone()
+                    }
+                })
+                .collect();
             if documentation.len() == 1 {
                 let text = &documentation[0];
                 documentation_arms
--- a/strum_macros/src/macros/strings/from_string.rs
+++ b/strum_macros/src/macros/strings/from_string.rs
@@ -3,7 +3,8 @@ use quote::quote;
 use syn::{Data, DeriveInput, Fields};
 
 use crate::helpers::{
-    non_enum_error, occurrence_error, HasStrumVariantProperties, HasTypeProperties,
+    non_enum_error, occurrence_error, HasInnerVariantProperties, HasStrumVariantProperties,
+    HasTypeProperties,
 };
 
 pub fn from_string_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
@@ -45,7 +46,6 @@ pub fn from_string_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
                     ))
                 }
             }
-
             default_kw = Some(kw);
             default = quote! {
                 ::core::result::Result::Ok(#name::#ident(s.into()))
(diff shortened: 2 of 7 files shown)


strum_macros/src/macros/enum_messages.rs
```diff
@@ -74,14 +74,17 @@ pub fn enum_message_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
...
             // Strip a single leading space from each documentation line.
-            let documentation: Vec<LitStr> = documentation.iter().map(|lit_str| {
-                let line = lit_str.value();
-                if line.starts_with(' ') {
-                    LitStr::new(&line.as_str()[1..], lit_str.span())
-                } else {
-                    lit_str.clone()
-                }
-            }).collect();
+            let documentation: Vec<LitStr> = documentation
```
ericmcbride, line 77: cargo fmt change. I can revert this if needed

strum_macros/src/macros/strings/from_string.rs
```diff
@@ -56,16 +56,34 @@ pub fn from_string_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
         let params = match &variant.fields {
             Fields::Unit => quote! {},
             Fields::Unnamed(fields) => {
-                let defaults =
-                    ::core::iter::repeat(quote!(Default::default())).take(fields.unnamed.len());
-                quote! { (#(#defaults),*) }
+                if let Some(ref value) = variant_properties.default_with {
```
ericmcbride, line 59: I wasn't if there was a new-type variant, if we wanted to be able to default each unnamed field.  It'll be easy enough to add this feature/tests

strum_macros/src/macros/strings/from_string.rs
```diff
@@ -56,16 +56,34 @@ pub fn from_string_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
...
             Fields::Named(fields) => {
-                let fields = fields
-                    .named
-                    .iter()
-                    .map(|field| field.ident.as_ref().unwrap());
-                quote! { {#(#fields: Default::default()),*} }
+                let mut defaults = vec![];
+                for field in &fields.named {
+                    let meta = field.get_variant_inner_properties()?;
+                    let field = field.ident.as_ref().unwrap();
```
ericmcbride, line 73: Wasn't sure about the unwrap here.  I don't like unwraps in general, and I could return an `ok_or_else` Syn error here.  I thought it'd just be weird if an ident on a field wasn't a thing.   Thoughts here?

strum_tests/tests/from_str.rs
```diff
@@ -15,6 +15,15 @@ enum Color {
     Purple,
     #[strum(serialize = "blk", serialize = "Black", ascii_case_insensitive)]
     Black,
+    Pink {
+        #[strum(default_with = "test_default")]
```
Peternator7, line 19: Let's add a test to make sure it works properly with a full module path

strum_macros/src/macros/strings/from_string.rs
Peternator7: If it's not too difficult, let's try to match the conventions that serde uses according to this doc: 

https://serde.rs/variant-attrs.html#deserialize_with


> `#[serde(deserialize_with = "path")]`
>
> Deserialize this variant using a function that is different from its implementation of Deserialize. The given function must be callable as `fn<'de, D>(D) -> Result<FIELDS, D::Error> where D: Deserializer<'de>`, although it may also be generic over the elements of FIELDS. Variants used with deserialize_with are not required be able to derive Deserialize.
> 
> FIELDS is a tuple of all fields of the variant. A unit variant will have () as its FIELDS type.

I'm not too worried about the generic bounds parts, however, if we can return a tuple, that's probably the clearest way to do it. Thoughts?

strum_macros/src/macros/strings/from_string.rs
Peternator7: I don't really mind unwrapping/expecting here. Maybe just use expect and tell people to file an issue against strum if they see this message

Peternator7 (commented):
Generally looks pretty good to me! Thanks for tackling this. I think it'll help a lot of use-cases

strum_macros/src/macros/strings/from_string.rs
ericmcbride: Let me see what I can do!

Peternator7:
Sorry for the long delay on this :( I don't have much time for this project anymore, but I'm trying to get through everything in the backlog. I think this is a cool addition to the project

PR #256, opened 2023-02-28
Add Train-Case case style

martinkauppinen:
`heck` 0.4.1 added this case style.


Peternator7:
lgtm 👍Thanks for the improvement

PR #257, opened 2023-03-05
Added enum_is

### LINKED ISSUE #252: is_* methods
I wrote a little derive macro that generates `is_*` methods for each enum variant, and I wonder if strum would be a good home for it. If so, I'll open a PR.

Example:
```rust
#[derive(Is)]
enum Foo {
    Bar {},
    Biz
}
```
The `Is` derive will add `is_bar()` and `is_biz()` methods, similar to `Option::is_some()`, `Option::is_none()`, `Result::is_ok()`, etc.

I think that:
```rust
if foo.is_bar() {
    ....
}
```
is a lot cleaner then:
```rust
if matches!(foo, Foo::Bar { .. }) {
    ....
}
```

Thoughts?

MendyBerger:
Resolves #252

### DIFF
--- a/strum_macros/Cargo.toml
+++ b/strum_macros/Cargo.toml
@@ -24,7 +24,6 @@ proc-macro2 = "1.0"
 quote = "1.0"
 rustversion = "1.0"
 syn = { version = "1.0", features = ["parsing", "extra-traits"] }
-convert_case = "0.6"
 
 [dev-dependencies]
 strum = "0.24"
--- a/strum_macros/Cargo.toml
+++ b/strum_macros/Cargo.toml
@@ -24,6 +24,7 @@ proc-macro2 = "1.0"
 quote = "1.0"
 rustversion = "1.0"
 syn = { version = "1.0", features = ["parsing", "extra-traits"] }
+convert_case = "0.6"
 
 [dev-dependencies]
 strum = "0.24"
--- a/strum_macros/src/macros/enum_is.rs
+++ b/strum_macros/src/macros/enum_is.rs
@@ -1,4 +1,4 @@
-use crate::helpers::non_enum_error;
+use crate::helpers::{non_enum_error, HasStrumVariantProperties};
 use heck::ToSnakeCase;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
@@ -14,10 +14,15 @@ pub fn enum_is_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
 
     let variants: Vec<_> = variants
         .iter()
-        .map(|variant| {
+        .filter_map(|variant| {
+            if variant.get_variant_properties().ok()?.disabled.is_some() {
+                return None;
+            }
+
             let variant_name = &variant.ident;
             let fn_name = format_ident!("is_{}", snakify(&variant_name.to_string()));
-            quote! {
+
+            Some(quote! {
                 #[must_use]
                 #[inline]
                 pub const fn #fn_name(&self) -> bool {
@@ -26,7 +31,7 @@ pub fn enum_is_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
                         _ => false
                     }
                 }
-            }
+            })
         })
         .collect();
 
--- a/strum_macros/src/macros/enum_is.rs
+++ b/strum_macros/src/macros/enum_is.rs
@@ -1,5 +1,5 @@
 use crate::helpers::non_enum_error;
-use convert_case::{Case, Casing};
+use heck::ToSnakeCase;
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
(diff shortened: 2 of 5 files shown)


strum_macros/Cargo.toml
```diff
@@ -24,6 +24,7 @@ proc-macro2 = "1.0"
 quote = "1.0"
 rustversion = "1.0"
 syn = { version = "1.0", features = ["parsing", "extra-traits"] }
+convert_case = "0.6"
```
Peternator7, line 27: convert_case has a dev_dependency on `strum`. Can we use `heck` to avoid pulling multiple casing libs?

strum_macros/src/macros/enum_is.rs
```diff
@@ -0,0 +1,60 @@
...
+                        }
+                    }
+                }
+                syn::Fields::Unnamed(values) => {
+                    let underscores = values.unnamed.iter().map(|_| quote! {_});
+                    quote! {
+                        #[must_use]
+                        #[inline]
+                        pub const fn #fn_name(&self) -> bool {
+                            matches!(self, &#enum_name::#variant_name (#(#underscores),*))
```
Peternator7, line 37: I think you can use `...` here instead of repeating `_`

strum_macros/src/macros/enum_is.rs
```diff
@@ -0,0 +1,60 @@
...
+            let variant_name = &variant.ident;
+            let fn_name = format_ident!("is_{}", variant_name.to_string().to_case(Case::Snake));
+
+            match &variant.fields {
+                syn::Fields::Named(_) => {
+                    quote! {
+                        #[must_use]
+                        #[inline]
+                        pub const fn #fn_name(&self) -> bool {
+                            matches!(self, &#enum_name::#variant_name { .. })
```
Peternator7, line 27: `matches!(..)` looks really good but it wasn't introduced until `1.42`. Could we actually generate to support a broader range of versions?

```rust
match self {
     &#enum_name::#variant_name { .. } => true,
      _ => false
}
```

Peternator7 (changes requested):
Looks really good! Just a couple of hopefully minor tweaks

strum_macros/src/macros/enum_is.rs
```diff
@@ -0,0 +1,60 @@
...
+
+pub fn enum_is_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
+    let variants = match &ast.data {
+        Data::Enum(v) => &v.variants,
+        _ => return Err(non_enum_error()),
+    };
+
+    let enum_name = &ast.ident;
+
+    let variants: Vec<_> = variants
```
Peternator7, line 15: strum supports a `disabled` attribute on enums. Should we support that for this derive as well?

strum_macros/src/macros/enum_is.rs
MendyBerger: You're right! You can use `{ .. }` for every type of enum variant, so I don't eventneed that match!

strum_macros/src/macros/enum_is.rs
MendyBerger: Done

strum_macros/Cargo.toml
MendyBerger: Seems like heck is handling numbers differently from convert-case, e.g.
heck:               "Number1Function" => "number1_function".
convert-case : "Number1Function" => "number_1_function".

Personally I prefer that numbers are treated as separate words like conver-case does, but can change it if you disagree.

See withoutboats/heck/issues/18

strum_macros/src/macros/enum_is.rs
MendyBerger: Don't see any docs for disabled. 

https://docs.rs/strum/latest/strum/?search=disabled returns no results.

Peternator7: See [this section](https://docs.rs/strum/latest/strum/additional_attributes/index.html#attributes-on-variants)

strum_macros/Cargo.toml
Peternator7: I agree that heck's behavior here is a little odd, but I'd prefer to not introduce a new dependency. Let's either, just do what heck does, or take heck's output and just insert a `_` before a number if it's preceded by a letter.

MendyBerger: Removed convert_case

strum_macros/src/macros/enum_is.rs
MendyBerger: Added disabled support

MendyBerger:
@Peternator7 I resolved all your comments

Peternator7:
Just release `0.25.0` which has this upgrade! Thanks for putting in the work on this new feature. It will provide a nice convenience for users.

PR #259, opened 2023-03-13
Adding support for int and bool EnumProperties

ejmount:
What it says on the tin. I was looking for this functionality for my own project and found stubs for it, so I've filled in the functionality and updated tests/docs accordingly. `cargo test` comes back all green after my changes.

### DIFF
--- a/strum/src/lib.rs
+++ b/strum/src/lib.rs
@@ -155,9 +155,9 @@ pub trait EnumMessage {
 
 /// `EnumProperty` is a trait that makes it possible to store additional information
 /// with enum variants. This trait is designed to be used with the macro of the same
-/// name in the `strum_macros` crate. Currently, the only string literals are supported
-/// in attributes, the other methods will be implemented as additional attribute types
-/// become stabilized.
+/// name in the `strum_macros` crate. Currently, string, integer and boolean literals
+/// are supported in attributes, the other methods will be implemented as additional
+/// attribute types become stabilized.
 ///
 /// # Example
 ///
@@ -171,7 +171,7 @@ pub trait EnumMessage {
 ///     #[strum(props(Teacher="Ms.Frizzle", Room="201"))]
 ///     History,
 ///     #[strum(props(Teacher="Mr.Smith"))]
-///     #[strum(props(Room="103"))]
+///     #[strum(props(Room=103))]
 ///     Mathematics,
 ///     #[strum(props(Time="2:30"))]
 ///     Science,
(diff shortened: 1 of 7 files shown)


strum/src/lib.rs
```diff
@@ -152,24 +152,22 @@ pub trait EnumMessage {
...
 /// assert_eq!("Ms.Frizzle", history.get_str("Teacher").unwrap());
+/// let maths = Class::Mathematics;
+/// assert_eq!(103, maths.get_int("Room").unwrap());
 /// ```
 pub trait EnumProperty {
     fn get_str(&self, prop: &str) -> Option<&'static str>;
-    fn get_int(&self, _prop: &str) -> Option<usize> {
-        Option::None
-    }
+    fn get_int(&self, prop: &str) -> Option<usize>;
```
Peternator7, line 168: `get_usize` is probably a more future-proof name

strum_macros/src/helpers/metadata.rs
```diff
@@ -242,15 +243,22 @@ impl Parse for VariantMeta {
...
 
 impl Parse for Prop {
     fn parse(input: ParseStream) -> syn::Result<Self> {
         use syn::ext::IdentExt;
 
         let k = Ident::parse_any(input)?;
         let _: Token![=] = input.parse()?;
-        let v = input.parse()?;
+        let v;
+        if input.peek(syn::LitBool) {
```
Peternator7, line 255: Can we implement `Parse` for `PropertyValue` then just `let v: PropertyValue = input.parse()?` ?

Peternator7 (changes requested):
This looks really great, and I know a couple have people have asked about this feature. Just a couple minor structure things and I think we can get this merged.

strum/src/lib.rs
ejmount: I'm agree and I'm happy to make that change, but I also thought it'd be worth mentioning that while it'll be more future proof, it's also technically a breaking change. I'd be _immensely_ surprised if any real code got broken, but I just wanted to check you're ok with that.

ejmount: I jinxed myself by saying earlier nothing would break, because there seems to be an odd dependency between `strum_macros` and a previous version of the main package. I've made the change and updated the Cargo.toml to use the local version, which solved the issue.

strum_macros/Cargo.toml
```diff
@@ -26,4 +26,4 @@ rustversion = "1.0"
 syn = { version = "1.0", features = ["parsing", "extra-traits"] }
 
 [dev-dependencies]
-strum = "0.24"
+strum = { path = "../strum" }
```
Peternator7, line 29: I don't believe you can publish to crates.io with a path dependency.

strum/src/lib.rs
Peternator7: Hey @ejmount, sorry for the delay getting back to you. I've been out of town for a few weeks. I missed that I had stubbed that method out at some point in the past. If you're still interested in completing this, here's what we can do. I can publish a new version of strum (not macros) with an updated trait definition then we can rebase your changes onto the new version.

ejmount: Yeah, it [sounds like](https://github.com/rust-lang/cargo/issues/1565) that's still not supported, so it'd be great if you could publish a new version to make that update.

ejmount:
So this came back to my attention recently so I've rebased it against the current release and resolved conflicts. I had another look at the dependency issue, and as far as I can tell everything works correctly if the name remains `get_int`, it's only trying to change the name that runs into compatibility issues. 

The version I've uploaded just now supports ints and bools and passes local tests, and IMO would be ready to merge if you're OK with `get_int` having a suboptimal name. (I'd think it very unlikely that the exact type of integer is going to be important for most use cases, so it wouldn't cause ambiguity in practice)

I'm not sure why the CI tests are failing on some platforms and not others, but the issue doesn't appear to relate to this change

Peternator7:
#391 ended up implementing this as well so I'm going to close out this PR. Sorry, I missed that you looped back around to this.

PR #261, opened 2023-03-26
Update syn to v2

CosmicHorrorDev:
This updates `syn` to v2. Changes involved:

- Removing the `Spanned` requirement from `get_metadata_inner()` since it no longer seems needed and `Spanned` now requires `ToTokens` to be implemented on the type too
- Changing `parse_terminated()`'s arguments to take the delimiter instead of passing it with a turbo-fish
- Switching `path` to the method for `Attribute`s
- Changing the structure of how literals are expressed for meta name-values
- Matching the old structure of how tokenstreams were represented for `Attribute`'s
  - I tried to minimize the changes required to match the old parsing. A `Meta::List(_)` holds an internal tokenstream which could be used, but it's the tokenstream inside the group instead of the containing token-group itself. Instead I opted for taking the surrounding tokenstream and discarding the leading path to get at the group

### DIFF
--- a/strum_macros/Cargo.toml
+++ b/strum_macros/Cargo.toml
@@ -23,7 +23,7 @@ heck = "0.4.1"
 proc-macro2 = "1.0"
 quote = "1.0"
 rustversion = "1.0"
-syn = { version = "1.0", features = ["parsing", "extra-traits"] }
+syn = { version = "2.0", features = ["parsing", "extra-traits"] }
 
 [dev-dependencies]
 strum = "0.24"
--- a/strum_macros/src/helpers/metadata.rs
+++ b/strum_macros/src/helpers/metadata.rs
@@ -1,12 +1,11 @@
-use proc_macro2::{Span, TokenStream};
+use proc_macro2::TokenStream;
 use syn::{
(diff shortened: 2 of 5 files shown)


CosmicHorrorDev:
`syn` bumped its MSRV up to 1.56, so it's expected that the builds back on 1.32 would fail

smoelius:
Can I please ask what the status of this PR is?

Peternator7:
The PR is fine from an implementation standpoint. However, I try to maintain two years of backwards compatibility, and rust 1.56 is from 10/2021. Is there an urgency in merging this change?

smoelius:
I wouldn't say it's urgent. It's more of a housekeeping matter. For one of my projects, this is the last dependency that still relies on syn 1.0.

But I can wait until October.

Thanks very much for your response.

Peternator7:
Just release `0.25.0` which has this upgrade! Decided it's not that far off from October and it's impractical to maintain a project that's on an unsupported syn version.

smoelius1:
Thank you!

martin-g:
Thank you!

Issue #132, opened 2020-10-07
ToString Keep/Rename/Remove


Peternator7:
Evaluating the naming scheme and value of all the macros used in strum. Some are more or less redundant such as `ToString` now that `Display` exists. 

This is a tracking issue to solicit feedback about the usefulness of `ToString` and whether it should be kept and if so, what it's name should be.

Peternator7:
Generally, `Display` should be preferred over `ToString`, but ToString has been kept around for compatibility. Given it's probably the wrong thing to do, I would propose removing this macro all together.

linclelinkpart5:
I'm in favor of this change, it would mirror what Rust's `std` is aiming for as well.

b-zee:
@Peternator7 Not sure if this is worth a separate issue, but there might be a reason to keep `ToString` around.

I assume it is not uncommon for people to use `strum_macros` in combination with `thiserror`. Using both leads to a conflict as both derives end up implementing `Display` (?):

```
#[derive(thiserror::Error, strum_macros::Display)]
enum Error {
    #[error("My custom error")]
    CustomError,
}
```

```
error[E0119]: conflicting implementations of trait `std::fmt::Display` for type `Error`
```

I assume `strum_macros::ToString` will be circumventing this problem, but it is deprecated.

PR #264, opened 2023-04-18
fix deprecation warning in additional_attributes example

rsnively:
The example code in [additional_attributes](https://docs.rs/strum/latest/strum/additional_attributes/index.html) brings up a deprecation warning related to issue #132 .

This change updates the example code to use `strum_macros::Display` instead of `strum_macros::ToString` (as well as removing some unneeded use statements).

Current example code with warning:
example_deprecation_warning

Fixed:
fixed

### DIFF
--- a/strum/src/additional_attributes.rs
+++ b/strum/src/additional_attributes.rs
@@ -20,11 +20,9 @@
 //!   - `Train-Case`
 //!
 //!   ```rust
-//!   use std::string::ToString;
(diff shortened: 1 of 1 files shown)


rsnively:
I think this was supposed to make it in with commit #180 but accidentally got overwritten in commit #182

Peternator7:
thanks!

PR #268, opened 2023-04-26
Fix EnumCount with disabled variants

### LINKED ISSUE #267: Disabled variant still included in Count
The [additional attributes docs](https://docs.rs/strum/latest/strum/additional_attributes/index.html#attributes-on-variants) define the `disabled` attribute as:

> `disabled`: removes variant from generated code.

However, when deriving the `EnumCount` trait, the disabled variant is still counted in the final count.

Here is some example code:

```rust
use strum::EnumCount;

#[derive(strum::EnumCount)]
enum Fields {
    Field0,
    Field1,
    Field2,
    #[strum(disabled)]
    Unknown,
}

fn main() {
    println!("Count: {}", Fields::COUNT);
}
````

Expected output: `Count: 3`
Actual output: `Count: 4`

This issue seems similar to #244

schultetwin1:
Do not include disabled variants in the count with EnumCount. Fixes #267

### DIFF
--- a/strum_macros/src/macros/enum_count.rs
+++ b/strum_macros/src/macros/enum_count.rs
@@ -2,11 +2,18 @@ use proc_macro2::TokenStream;
 use quote::quote;
 use syn::{Data, DeriveInput};
 
+use crate::helpers::variant_props::HasStrumVariantProperties;
 use crate::helpers::{non_enum_error, HasTypeProperties};
 
 pub(crate) fn enum_count_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
     let n = match &ast.data {
-        Data::Enum(v) => v.variants.len(),
+        Data::Enum(v) => v.variants.iter().try_fold(0usize, |acc, v| {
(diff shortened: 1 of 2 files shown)


Peternator7:
Thanks for the PR! I'm in favor of this change, but it's a breaking change so I just need to figure out what release to bundle it into.

schultetwin1:
Not a problem! Sounds good!

Peternator7:
Thanks for continuing to think about this. Let's actually remove the feature flag and I'm going to put a couple of breaking changes into `0.25`.

schultetwin1:
> Thanks for continuing to think about this. Let's actually remove the feature flag and I'm going to put a couple of breaking changes into `0.25`.

Sounds good! Done!

Peternator7 (approved):

Peternator7:
Just release `0.25.0` which has this upgrade! Thanks for putting the work in :)

schultetwin1:
Amazing, thank you!!

PR #271, opened 2023-05-16
ci: add github workflows

joshka:
This PR adds Github CI workflows to replace appveyor.

[@jonhoo's ci conf repo](https://github.com/jonhoo/rust-ci-conf/) has
a reasonable stub of CI workflows and configuration
for rust crates. It includes config for dependabot, codecov.io, and
workflows for checking format and clippy, testing on multiple OSes,
determining the minimum supported rust version, checking for problems
with minimal versions, and runs scheduled tasks to ensure that new rust
versions and dependencies do not break this crate.

This is a merge using the --allow-unrelated-histories flag, that brings
in the entire history of that repo. This enables the CI to be later
merged if configuration changes are made upstream.

I've got a PR out against the upstream repo that adds a bit more documentation
about each workflow
<https://github.com/jonhoo/rust-ci-conf/pull/10>

## Review Approach
When reviewing this change, there's two parts to consider - the history pulled in from the upstream repo which should probably not be squashed (and which contains the bulk of the implementation.
- https://github.com/jonhoo/rust-ci-conf/commits/main

And then the changes specifically for strum which probably can be squashed into a single commit (though I have not yet done so as the commit history is relevant to understanding why specific choices were made in configuring CI
- https://github.com/joshka/strum/compare/d668c3d314ede430deaab41e9248afbff73172da...joshka:strum:ci

## Concerns:
- 1.32.0 doesn't seem to build on MacOS right now.
- There is a fix for building no-std correctly (building strum_nostd_tests and strum together adds the std feature, for some strange reason without the change)
- The appveyor config is a lot more granular around the build targets - I wasn't sure if this was necessary or whether the OS toolchain based approach was sufficient. Happy to fix this if needed.

## Testing

I ran this pull request in my own fork at https://github.com/joshka/strum/pull/1/commits with success (except for the 1.32.0 issue on Macs)


joshka:
Closing this out as too complex to solve the problem here. Will make a clearer shorter PR that is easier to consume for CI.

PR #276, opened 2023-06-23
Handle rustoc comments in `#[derive(FromRepr)]`

### LINKED ISSUE #275: derive(FromRepr) broken in 0.25.0
image

Worked flawlessly with 0.24.1

smoelius:
I think this fixes #275, but please scrutinize.

### DIFF
--- a/strum_macros/src/macros/from_repr.rs
+++ b/strum_macros/src/macros/from_repr.rs
@@ -16,7 +16,15 @@ pub fn from_repr_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
     for attr in attrs {
         let path = attr.path();
 
-        let mut ts = attr.meta.require_list()?.to_token_stream().into_iter();
+        let mut ts = if let Ok(ts) = attr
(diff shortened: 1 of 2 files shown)


smoelius:
It looks like the same failures occurred on master, so I assume they are not caused by this change: https://ci.appveyor.com/project/Peternator7/strum/builds/47333125

Please let me know if that is not correct.

Peternator7:
Hey @smoelius, thanks for tackling this. I'll take a closer look at the logic, but what immediately jumps out is that `let-else` is too new a feature for strum. It's great for brevity for sure, but for back-compat, we'll need to use an `if let {} else {}` to not break older rust versions.

I also have a PR to fix the CI to the proper msrv. After that merges, you should be able to rely on the builds again. Sorry about that.

smoelius:
> `let-else` is too new a feature for strum.

:facepalm: This should have occurred to me, given https://github.com/Peternator7/strum/pull/261#issuecomment-1537230542.

Issue #272, opened 2023-05-19
Enum Map derive macro


PokeJofeJr4th:
I think it would be useful to have a map that you can index with an enum. I had the idea while working on a game inventory system where items are enums. I've started [my own fork](https://github.com/PokeJofeJr4th/strum) to start the macro. I'm struggling with the dependency on EnumIter and possible support for variants with attached data.

I still have a long way to go before it's PR-quality. I'm posting here to gauge whether my idea seems worthwhile and to ask for help. I'm very new to open source so I'm not sure if I'm doing all this right.

itsjunetime:
I also think this would be really useful. One could use a hashmap but then you'd have to `.unwrap()` every time you `.get` a value even if you know that every enum variant has a valid value, and being able to make a custom struct for each enum to act as a variant map would have a lot of uses.

I created an initial implementation of a derive macro for this on [my own repo here](https://github.com/itsjunetime/enum_map). If you have the following code:
```rust
#[derive(EnumMap)]
enum Size {
    Small,
    Medium,
    Large,
}
```
it expands to:
```rust
struct SizeMap<T> {
    small: T,
    medium: T,
    large: T,
}
impl<T> SizeMap<T> {
    fn new(small: T, medium: T, large: T) -> SizeMap<T> {
        SizeMap { small, medium, large }
    }
    fn get(&self, variant: Size) -> &T {
        match variant {
            Size::Small => &self.small,
            Size::Medium => &self.medium,
            Size::Large => &self.large,
        }
    }
    fn set(&mut self, variant: Size, new_value: T) {
        match variant {
            Size::Small => self.small = new_value,
            Size::Medium => self.medium = new_value,
            Size::Large => self.large = new_value,
        }
    }
}
```
I'd be happy to turn this into a PR if the owners of this repo would be interesting in having this macro. I also feel that there are questions to be discussed before releasing something like this, e.g.:
1. How do we specify what the visibility of the expanded struct could be? It would be useful to allow it to sometimes be `pub` but it can't always be `pub` since it may then leak an external type if the original enum is not `pub`
2. Should we have other initializers? Having some other options such as setting the value of all variants with a `Copy` value, or setting the value of all variants with a provided closure would be very useful, I think.
3. How do we deal with variants that have data (as mentioned above)? Should we even try to accommodate for that? We could perhaps look to the policy of `EnumIter` for this one, since it would run into a similar problem.

[...]

PR #279, opened 2023-06-26
Add Enum Map Derive Macro

PokeJofeJr4th:
This started as Issue #272 
```rust
#[derive(EnumMap)]
enum MyEnum {
  VariantName,*
}
```
This macro creates a new type `MyEnumMap<T>` with a field of type `T` for each variant of `MyEnum`. It also adds the following:
- Automatic derive for Debug, Clone, Default, PartialEq, Eq, Hash
- Implement `Index<T>` and `IndexMut<T>`
- `fn new(variant_name: T,*) -> MyEnumMap<T>` constructor to specify each struct field
- `fn from_closure(func: Fn(MyEnum)->T) -> MyEnumMap<T>` constructor to initialize each struct field using a function
- `fn transform(&self: MyEnumMap<T>, func: Fn(MyEnum,&T) -> U) -> MyEnumMap<U>` method to create a new map by applying a function to each field
- if `T` implements `Clone`, generate `fn filled(value: T) -> MyEnumMap<T>` constructor to fill each struct field with the given vaalue
- `fn all(self: MyEnumMap<Option<T>>) -> Option<MyEnumMap<T>>` checks if all fields are `Some`
- `fn all_ok(self: MyEnumMap<Result<T, E>>) -> Option<MyEnumMap<T>>` checks if all fields are `Ok`
- variants marked with `#[strum(disabled)]` are not included as struct fields and panic when passed as an index.

Questions:
- Should `EnumMap` be renamed to clarify that it doesn't have many features of conventional maps?
- Are the `all` and `all_ok` functions helpful? Are they within the scope of Strum?
- I'm currently disallowing extra data on variants. Should I just ignore it and use the discriminant? Should I treat the underlying data as a key to a more conventional map?
- Are there any other traits/functions I'm forgetting about that should be included?


PokeJofeJr4th:
the build is only failing because the version of rust they're using is out of date

strum_macros/src/macros/enum_map.rs
```diff
@@ -0,0 +1,217 @@
...
+        return Err(syn::Error::new(
+            Span::call_site(),
+            "This macro doesn't support enums with lifetimes.",
+        ));
+    }
+
+    let Data::Enum(data_enum) = &ast.data else {
+        return Err(non_enum_error())
+    };
+    let map_name = syn::parse_str::<Ident>(&format!("{}Map", name)).unwrap();
```
itsjunetime, line 23: Is there a reason that you didn't use `format_ident!` here instead? It would avoid an unwrap and (imo) be cleaner.

strum_macros/src/macros/enum_map.rs
```diff
@@ -0,0 +1,217 @@
...
+    // struct fields of the form `variant: func(MyEnum::Variant, self.variant),`
+    let mut transform_fields = Vec::new();
+
+    // identifiers for disabled variants
+    let mut disabled_variants = Vec::new();
+    // match arms for disabled variants
+    let mut disabled_matches = Vec::new();
+
+    for variant in variants {
+        // skip disabled variants
```
itsjunetime, line 48: Having a variant that could be disabled and cause a panic if indexed with makes me nervous, but it seems to be in line with what this crate normally does. I feel like it would make more sense, though, to ignore the disabled attribute for this specific macro (among others, like `AsRefStr`) since generating code for the ignored attributes can't cause any harm.

If they already weren't going to index into this map with that variant (e.g. they were using `EnumIter` with a disabled variant to iterate through all the variants and modify this `EnumMap` with each variant), then there's no harm in not panicking if they try to index with what would be a 'disabled' variant. It could cause confusing behavior if they accidentally do index into this with the disabled variant (since it would basically do nothing instead of panicking), but I feel like that's much better behavior than panicking. Perhaps this comment should be made into an issue instead to discuss there.

strum_macros/src/macros/enum_map.rs
```diff
@@ -0,0 +1,217 @@
...
+        if variant.get_variant_properties()?.disabled.is_some() {
+            let disabled_ident = &variant.ident;
+            let panic_message =
+                format!("Can't use `{disabled_ident}` with `{map_name}` - variant is disabled for Strum features");
+            disabled_variants.push(disabled_ident);
+            disabled_matches.push(quote!(#name::#disabled_ident => panic!(#panic_message),));
+            continue;
+        }
+
+        // Error on fields with data
```
itsjunetime, line 58: I agree with this decision - calling this a `map` implies that every possible variant has its own value and if you have a variant that has data (e.g. `Name(&'static str)`) it could potentially be confusing if indexing with that variant but different data (e.g. `map[Name("John")]` vs `map[Name("Jane")]`) returns/modifies the same value.

However, this doesn't exactly seem to go in line with what this crate normally does - e.g. with `EnumIter`, it just generates `default::Default()` for each field if a variant has data (instead of refusing to expand the macro). If we wanted to keep behavior similar to the other derive macros here, we might want to still expand the macro but just ignore all data on each variant and only match on variant type. I can't think of a specific use-case where allowing data would be necessary, so I'm still in favor of keeping this as-is, but I'm not quite certain what other contributors would be in favor of.

strum_macros/src/macros/enum_map.rs
```diff
@@ -0,0 +1,217 @@
...
+        let Fields::Unit = &variant.fields else {
+            return Err(syn::Error::new(
+                variant.fields.span(),
+                "This macro doesn't support enums with non-unit variants",
+            ))
+        };
+
+        let pascal_case = &variant.ident;
+        // switch PascalCase to snake_case. This naively assumes they use PascalCase
+        let snake_case = format_ident!(
```
itsjunetime, line 68: Perhaps for this code we could reuse `enum_is.rs`'s `snakify` since that already accomplishes the purpose?

itsjunetime (approved):
I like it overall. There are a few things that I think could be changed for the better, but I'm happy with it as is and would love to see it merged.

strum_macros/src/macros/enum_map.rs
PokeJofeJr4th: No, I didn't realize that was a thing. I thought I took that from EnumIter though.

strum_macros/src/macros/enum_map.rs
PokeJofeJr4th: I think that unexpected behavior would be more detrimental to a project than a panic, since the panic is obvious what the problem is

strum_macros/src/macros/enum_map.rs
PokeJofeJr4th: Yes, I just copied that from you

itsjunetime:
In response to some of your other questions which I've been thinking about:
1. I'm not quite certain what a better name would be. I guess if there was some way to specify that it's a map that has `Some` value for every possible key, that could be more clear, but I don't know what that would be.
2. I feel like the `all_ok` function is definitely useful, but would be better if it returned a `Result<T, E>` where it returns `Err` with the first error encountered while looking through the variants (plus, this would also make it really simple to implement - just `EnumMap { #(variant: variant?,)* }` in the body). I was previously going to say that `all` wouldn't be super useful, but I was thinking about my use cases and realized it could actually be really useful. E.g. I would like to store data in an `EnumMap` to disk, and the library I'm using for that can store a `HashMap` right out of the box. So I plan to convert the `EnumMap` to a `HashMap` when storing like so:

    ```rust
    let mut map = HashMap::new();
    for var in Enum::iter() {
        map[var] = enum_map[var]
    }
    ```
    But to convert back after reading the `HashMap` from the disk, The only safe and clean way I can think of (besides manually getting the value for each variant) would be:
    ```rust
    let Some(enum_map) = EnumMap::from_closure(|v| hash_map.get(v)).all() else {
        println!("oops!");
    };
    ```
    So actually, I feel like `all` could definitely be useful if you need to convert from other similar storage mediums.
4. I put my thoughts about non-unit variants in another comment
5. I think `impl`ing `Clone` where `T: Clone` could be nice. I guess if you wanted to work around it, you could just do `EnumMap::from_closure(|v| other_map[v].clone())`, but just doing `other_map.clone()` would be cleaner and clearer a bit.

PokeJofeJr4th:
1. ChatGPT's best suggestion was `MyEnumTable`, which I kinda like
2. I did change the `all_ok` to what you suggested. However, for your exact case, would it be better to implement a way to directly store the EnumMap?
3. yeah that's still a hard one
4. That's already a thing with the `#[derive(...)]` thingy; I've added a test to clarify that

itsjunetime:
1. Yeah, I think calling it a table could be good; if people would prefer that, then I have no objections
2. It would definitely be better for me to directly store the `EnumMap`, but with the library that I'm using, that requires `derive`ing a bespoke macro, which I don't think I can do right now for the generated `EnumMap`. However, with that in mind, perhaps we could add an attribute like `#[enum_map(derive = "MyTrait, MyOtherTrait")]` so that you can add other traits to the generated struct's `derive` attribute? I'm not attached to that syntax, just the general idea of adding extra traits to derive.

And thanks for pointing out that it already derives `Clone`; I must've missed that.

PokeJofeJr4th:
found a significant issue
https://github.com/PokeJofeJr4th/strum/issues/8

Peternator7:
Could you clarify what the significant issue was? It's good to consider that someone could use a reserved word; can we fix that by prepending an underscore to their identifier?

PokeJofeJr4th:
Sorry, I see now that you have a perfectly fine fix (prepending `_`). I previously thought we'd have to figure out if it was a reserved keyword and use `r#` to escape it; but since it's not publicly facing the underscore solution is perfectly valid

strum_macros/src/macros/enum_table.rs
```diff
@@ -0,0 +1,193 @@
...
+    let mut doc_comment = format!("A map over the variants of `{name}`");
+
+    if gen.lifetimes().count() > 0 {
+        return Err(syn::Error::new(
+            Span::call_site(),
+            "`EnumTable` doesn't support enums with lifetimes.",
+        ));
+    }
+
+    let Data::Enum(data_enum) = &ast.data else {
```
Peternator7, line 20: let-else isn't supported in 1.56

strum_macros/src/macros/enum_table.rs
```diff
@@ -0,0 +1,193 @@
...
+            let disabled_ident = &variant.ident;
+            let panic_message =
+                format!("Can't use `{disabled_ident}` with `{table_name}` - variant is disabled for Strum features");
+            disabled_variants.push(disabled_ident);
+            disabled_matches.push(quote!(#name::#disabled_ident => panic!(#panic_message),));
+            continue;
+        }
+
+        // Error on variants with data
+        let Fields::Unit = &variant.fields else {
```
Peternator7, line 59: let-else

strum_macros/src/macros/enum_table.rs
```diff
@@ -0,0 +1,193 @@
...
+        set_matches.push(quote! {#name::#pascal_case => self.#snake_case = new_value,});
+        closure_fields.push(quote! {#snake_case: func(#name::#pascal_case),});
+        transform_fields.push(quote! {#snake_case: func(#name::#pascal_case, &self.#snake_case),});
+        pascal_idents.push(pascal_case);
+        snake_idents.push(snake_case);
+    }
+
+    // Error on empty enums
+    if pascal_idents.is_empty() {
+        return Err(syn::Error::new(
```
Peternator7, line 80: Maybe I missed something. Is this strictly necessary?

strum_macros/src/macros/enum_table.rs
```diff
@@ -0,0 +1,193 @@
...
+        impl<T: Clone> #table_name<T> {
+            #[doc = #doc_filled]
+            #vis fn filled(value: T) -> #table_name<T> {
+                #table_name {
+                    #(#snake_idents: value.clone(),)*
+                }
+            }
+        }
+
+        impl<T> #table_name<T> {
```
Peternator7, line 123: Can we add a `get` and `get_mut` methods to compliment Index and match the semantics of arrays, Vecs and HashMaps?

strum_macros/src/macros/enum_table.rs
```diff
@@ -0,0 +1,193 @@
...
+            #[doc = #doc_transform]
+            #vis fn transform<U, F: Fn(#name, &T)->U>(&self, func: F) -> #table_name<U> {
+              #table_name {
+                #(#transform_fields)*
+              }
+            }
+
+        }
+
+        impl<T> core::ops::Index<#name> for #table_name<T> {
```
Peternator7, line 149: I suggest doing `::core::ops::Index` to ensure we always pull this method from `core` in case someone shadowed the module.

strum_macros/src/macros/enum_table.rs
```diff
@@ -0,0 +1,193 @@
...
+        impl<T, E> #table_name<Result<T, E>> {
+            #[doc = #doc_result_all_ok]
+            #vis fn all_ok(self) -> Result<#table_name<T>, E> {
+                Ok(#table_name {
+                    #(#snake_idents: self.#snake_idents?,)*
+                })
+            }
+        }
+    })
+}
```
Peternator7, line 193: What's the plan for `Iterator`?

strum_macros/src/lib.rs
```diff
@@ -409,6 +409,45 @@ pub fn enum_is(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
+///
+/// let mut complex_map = ColorTable::from_closure(|color| match color {
+///     Color::Red => 0,
+///     _ => 3
+/// });
+/// complex_map[Color::Green] = complex_map[Color::Red];
+/// assert_eq!(complex_map, ColorTable::new(0, 3, 0, 3));
+///
+/// ```
+#[proc_macro_derive(EnumTable, attributes(strum))]
```
Peternator7, line 441: I don't feel too strongly, but rust tends to use the "Map" terminology rather than "Table." Thoughts about calling this `EnumMap` instead?

*I thoughts that's what the feature would be called based on the PR title.

Peternator7 (changes requested):
Thanks for the work on this PR. I've been busy this summer and haven't had a lot of time to review changes, but I think it's a reasonable feature to include.

strum_macros/src/lib.rs
PokeJofeJr4th: We had a conversation about this a while back and the conclusion was that `Map`s tend to be dynamically sized based on how full they are, whereas this type always has a value for each key. We thought these differences were enough that calling the structure a `Map` would cause people to make incorrect assumptions about how it works. That said, I don't feel too strongly about it either.

strum_macros/src/macros/enum_table.rs
PokeJofeJr4th: I don't believe we need these. The main reason to have both `Index` and `get` for existing types is that a key doesn't always map to a value. For `EnumTable`, those methods wouldn't be any different from just using `Index` and `IndexMut`. Again, I don't care much on this, just want to make sure we're on the same page.

strum_macros/src/macros/enum_table.rs
PokeJofeJr4th: It's not *strictly* necessary, but I think it's a good idea to include. Otherwise an empty enum produces this error, which tells the developer a lot less about what they're doing wrong.
```
error[E0392]: parameter `T` is never used
  --> strum_tests\tests\enum_table.rs:22:10
   |
22 | #[derive(EnumTable)]
   |          ^^^^^^^^^ unused parameter
   |
   = help: consider removing `T`, referring to it in a field, or using a marker such as `PhantomData`
   = help: if you intended `T` to be a const parameter, use `const T: usize` instead
   = note: this error originates in the derive macro `EnumTable` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0392`.
error: could not compile `strum_tests` (test "enum_table") due to previous error
```
The only other valid option I see is using phantomdata, and I don't think it makes sense to allow someone to derive a data structure that can't hold any data.

strum_macros/src/macros/enum_table.rs
PokeJofeJr4th: I think it makes sense to just use `MyEnum::iter().map(|variant| my_enum_map[variant])`. I don't know how to implement it otherwise without basically rewriting all of `EnumIter`.

itsjunetime:
I feel like it would be really useful to add custom derive commands for each struct this is used on, e.g.
```rust
#[derive(EnumTable)]
#[strum(table_derives = "Serialize, Deserialize")]
enum Color {
...
}
```
this way, `ColorTable` would have `#[derive(Serialize, Deserialize)]` added on its declaration since you can't do that otherwise (at least, as far as I know). Would anyone be opposed to this?

PokeJofeJr4th:
I like that but will the namespaces work on it? Would the user having macros like Serialize and Deserialize in the same scope as `#[derive(EnumTable)]` be enough?

Peternator7:
Published [strum](https://crates.io/crates/strum) version `0.26.0` that includes this feature. I've labeled it experimental because I'm not totally sure the api surface won't change, but think it's an interesting start. Thanks for the work on this and sorry for the delay merging it.

Issue #277, opened 2023-06-25
Unwrap methods


MendyBerger:
In a similar spirit to #252, would it be helpful to add `unwrap` methods to enums for variants?

E.g. 
```rust
enum Foo {
    Simple,
    Single(bool),
    Double(i32, String),
    Named {
        a: u8,
        b: u16,
    }
}
```
We could add the following methods:
```rust
impl Foo {
    pub fn unwrap_single(self) -> bool;
    pub fn unwrap_double(self) -> (i32, String);
    
    // we can also add reference versions of these (because we can't assume that .as_ref() is implemented on Foo)
    pub fn unwrap_single_ref(&self) -> &bool;
    pub fn unwrap_double_ref(&self) -> (&i32, &String);
}
```
Adding unwrap to `Foo::Simple` doesn't make much sense as there's nothing to unwrap.
Adding unwrap to `Foo::Named` is not possible as we'd need a struct name to return, an we don't have one. (unless we add a new struct `FooNamed` with the same values as `Foo::Named`, but that would probably be a bit too much.)
But for variants with unnamed values it might be useful.

We could also add `unwrap_or`, `unwrap_or_else`, `unwrap_unchecked` methods.

PokeJofeJr4th:
I wrote the rest of the comment before looking at the `is_*` methods. Given that those were ultimately added to Strum, I think these work well. I'll start a fork.

This is definitely possible, but I'm not sure how useful it would be. Consider the following pattern:
```rust
// vanilla rust
let Foo::Double(integer, string) = foo else { panic!() };
// using unwraps
let (integer, string) = foo.unwrap_double();

// vanilla rust
let Foo::Named{a, b} = foo else { panic!() };
// no way to do this with these unwraps
```

*I assume you meant that adding unwrap to `Foo::Simple` wouldn't make sense?*

MendyBerger:
> I assume you meant that adding unwrap to Foo::Simple wouldn't make sense?

Good catch! I fixed my original comment.

PokeJofeJr4th:
I also just had another idea that I'd like to be able to derive: a version of the `Result::ok` method for each variant to get an option.
```rust
impl Foo {
  pub fn single(&self) -> Option<bool> {
    match self {
      Self::Single(b) => Some(b),
      _ => None,
    }
  }
}
```
This would be useful in map methods, since you could pull all the variants out like this:
```rust
let vec: Vec<Foo> = vec![...];
[...]

PR #285, opened 2023-07-17
Added EnumTryAs

MendyBerger:
Resolves https://github.com/Peternator7/strum/issues/277

Didn't do unwrap, mentioned reasoning in this comment. Let me know if you disagree. https://github.com/Peternator7/strum/issues/277#issuecomment-1625989390


strum_macros/src/macros/enum_try_as.rs
```diff
@@ -0,0 +1,80 @@
...
+                        pub const fn #ref_fn_name(&self) -> Option<(#(&#types),*)> {
+                            match self {
+                                #enum_name::#variant_name (#(#field_names),*) => Some((#(#field_names),*)),
+                                _ => None
+                            }
+                        }
+
+                        #[must_use]
+                        #[inline]
+                        pub fn #mut_fn_name(&mut self) -> Option<(#(&mut #types),*)> {
```
Peternator7, line 58: Same here, `::core::option::Option`

strum_tests/tests/enum_try_as.rs
```diff
@@ -0,0 +1,44 @@
+use strum::EnumTryAs;
+
+#[derive(EnumTryAs)]
+enum Foo {
```
Peternator7, line 4: Let's add a unit like variant and a c-struct like variant to make sure we cover all the cases.

Peternator7 (changes requested):

Peternator7:
Really good start imo; I think just doing the "try_as" option methods provides the right balance between exposing new functionality via macro and steering users back towards existing concepts such as option.

PR #288, opened 2023-08-01
EnumDiscriminant  inherits the repr and discriminant values

### LINKED ISSUE #283: EnumDiscriminants should inherit the repr of the enum they are derived from
The below example code will generate `VehicleDiscriminants` enum with `Car` and `Truck` variants but it will be `repr(usize)` instead of inheriting the `u8` repr of `Vehicle` additionally they will have the default enum representations (`Car = 0` and `Vehicle = 1` in this case)

```
// Custom discriminant tests
#[derive(EnumDiscriminants, Debug, PartialEq)]
#[strum_discriminants(derive(FromRepr))]
#[repr(u8)]
enum Vehicle {
    Car(CarModel) = 1,
    Truck(TruckModel) = 3,
}
```

I would have expected the above code to produce the following:
```
#[repr(u8)]
#[derive(FromRepr)]
enum VehicleDiscriminants {
    Car = 1,
    Truck = 3,
}
```

jscatena88:
Would close #283

Since `FromRepr` also uses the `repr` attribute as well, I  attempted to centralize the parsing to retrieve it into `helpers/type_props.rs`. 

I don't have any prior experience working with proc macros or `syn` in general so feel free to give stylistic pointers in addition to feedback on the functionality itself

### DIFF
--- a/strum_macros/src/helpers/type_props.rs
+++ b/strum_macros/src/helpers/type_props.rs
@@ -21,6 +21,7 @@ pub struct StrumTypeProperties {
     pub discriminant_others: Vec<TokenStream>,
     pub discriminant_vis: Option<Visibility>,
     pub use_phf: bool,
+    pub enum_repr: Option<TokenStream>,
 }
 
 impl HasTypeProperties for DeriveInput {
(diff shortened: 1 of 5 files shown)


Peternator7:
hey @jscatena88, sorry for the delay. It looks great! thanks for the update. It's a breaking change, but more consistent so let's do it.

Peternator7:
Published [strum](https://crates.io/crates/strum) version `0.26.0` that includes this feature.

PR #289, opened 2023-08-03
Enable doc-test for "strum_discriminants(vis(pub))"

hasezoey:
This PR enables a previously disabled doc-test, i locally tested it and it tests fine and currently on [docs.rs](https://docs.rs/strum/0.25.0/strum/derive.EnumDiscriminants.html) it shows that code completely without syntax highlighting, making the code harder to read than necessary


Peternator7:
lgtm, thanks!

PR #292, opened 2023-08-27
feat: added discriminator fn

DevSlashRichie:
I've just added a `discriminant()` function which from the original enum returns it matched discriminant. 
Is just less verbose.


Peternator7:
Hey @DevSlashRichie, thanks for opening a PR here. Just to make sure I'm understanding the intention correctly, the purpose of this PR is to add an inherent method because it's shorter than using the Trait impl? 

If so, I appreciate the consideration, but I don't think it's a good fit. As convenient as inherent methods are, they are always in scope so it can cause annoying collisions with other userd-defined methods.

PR #293, opened 2023-08-28
fix EnumIter for enums with const generic

PingPongun:
Currently, EnumIter for struct with constant generic results in error E0747 (constant provided when a type was expected).
That's because generated iter struct looks sth like:
```Rust
struct LanguageIter<LAUNCHER> {
        idx: usize,
        back_idx: usize,
        marker: ::core::marker::PhantomData<()>,
    }
```
for enum:
```Rust
#[derive( EnumIter)]
enum Language<const LAUNCHER: bool> {...}
```
Which is ok for ordinary(type) generics, but for const generics iter struct definition has to look like:
```Rust
struct LanguageIter<const LAUNCHER: bool> {...}
```
this PR makes EnumIter work with const generics, by using $impl_generics instead of $ty_generics for struct definition.
Also added const generic to enum in tests.

### DIFF
--- a/strum_macros/src/macros/enum_iter.rs
+++ b/strum_macros/src/macros/enum_iter.rs
@@ -74,7 +74,7 @@ pub fn enum_iter_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
         #[allow(
             missing_copy_implementations,
         )]
-        #vis struct #iter_name #ty_generics {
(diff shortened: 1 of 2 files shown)


Peternator7 (approved):

Peternator7:
Looks good. Thanks for the fix :)

PR #294, opened 2023-09-06
Use enum named fields in to_string macro 

Arian8j2:
Not much, Just see the example:
```rust
#[derive(strum_macros::Display)]
enum Color {
    #[strum(to_string = "saturation is {sat}")]
    Purple { sat: usize },
}

assert_eq!(
    String::from("saturation is 10"),
    (Color::Purple { sat: 10 }).to_string().as_ref()
);

```

### DIFF
--- a/strum_macros/src/lib.rs
+++ b/strum_macros/src/lib.rs
@@ -317,6 +317,10 @@ pub fn to_string(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
 ///     },
 ///     Blue(usize),
 ///     Yellow,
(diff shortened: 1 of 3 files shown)


Arian8j2:
Seems like environments in **CI** that failed uses old version of `rustc` and doesn't support format like these:
```rust
let name = "alex";
format!("my name is {name}")
```
and formats like these are used internally to implement this pr,  but i think explicitly assigning each variable like these:
```rust
let name = "alex";
format!("my name is {name}", name = name)
```
will make this possible in older rustc versions, I will try to implement this tomorrow.

PR #296, opened 2023-09-15
Add the `prefix` attribute

chanman3388:
The idea here was to allow the addition of the `prefix` attribute such that when displaying the enum variant that said variant will have the prefix. This should be exempt from other formatting. Example usage:

```
use strum::{Display, EnumString};

#[derive(Debug, EnumString, Display)]
#[strum(prefix = "prefix")]
enum Thing {
    Test,
}

fn main() {
    println!("{}", Thing::Test);  // prints "prefixTest"
}
```

Could possibly add some kind of skipping mechanism for individual variants.

### DIFF
--- a/strum_macros/src/helpers/case_style.rs
+++ b/strum_macros/src/helpers/case_style.rs
@@ -1,5 +1,6 @@
 use heck::{
-    ToKebabCase, ToLowerCamelCase, ToShoutySnakeCase, ToSnakeCase, ToTitleCase, ToUpperCamelCase, ToTrainCase,
+    ToKebabCase, ToLowerCamelCase, ToShoutySnakeCase, ToSnakeCase, ToTitleCase, ToTrainCase,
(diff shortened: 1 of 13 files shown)


Peternator7:
Looks great! Merged

cheban96:
> The idea here was to allow the addition of the `prefix` attribute such that when displaying the enum variant that said variant will have the prefix. This should be exempt from other formatting. Example usage:
> 
> ```
> use strum::{Display, EnumString};
> 
> #[derive(Debug, EnumString, Display)]
> #[strum(prefix = "prefix")]
> enum Thing {
>     Test,
> }
> 
> fn main() {
>     println!("{}", Thing::Test);  // prints "prefixTest"
> }
> ```
> 
> Could possibly add some kind of skipping mechanism for individual variants.

It will be amazing if that feature support prefix for fields

Issue #198, opened 2021-12-01
Generation of associated const array of sized EnumCount


ratmice:
This is just a (perhaps silly) thought, it is pretty easy to collect into a vec or slice I suppose, with
`let foo: Vec<Foo> = Foo::iter().collect();`

But it occurred to me that because of the non-const trait functions on iterator you cannot 
`let foo: [Foo; Foo::COUNT]; = Foo::iter().collect();`

It might be nice if there was a way to generate an associated const array, perhaps named `Foo::ITEMS` such that essentially the existing `EnumIter` for `Foo::iter()` would be `Foo::ITEMS.iter()`

As I said, i'm not doing anything which needs it, but similar to #96 where the `fn count()` was changed to be an associated const.

ratmice:
Of course, I just realized that since `Default::default()` isn't const this probably be feasible...

Peternator7:
Alas, you are correct, without const trait methods, this would be difficult to do in the general case. 

I'm going to leave it open because I like the idea, but I don't think we can really do anything here without rustc support.

dzmitry-lahoda:
seems we still have no const traits?

would generating without traits work?

PR #297, opened 2023-09-21
Add `StaticVariantsArray` trait and derive

Syndelis:
## About this PR

This PR introduces a new trait `StaticVariantsArray` which defines an associated constant `ALL_VARIANTS: &'static [Self]`. Accompained to it is a derive that will create this array from an `enum` composed of only unit-type variants.

## Motivation

While working in a UI project with [ImGui](https://github.com/imgui-rs/imgui-rs), I stumbled upon [this method, Ui::combo](https://docs.rs/imgui/latest/imgui/struct.Ui.html#method.combo), which expects a slice and a mapping function to build the combobox's options. My options, as you might've guessed, are defined in an enumerator, which looks a little like this:

```rust
enum Operation {
  Add,
  Sub,
  Mul,
  Div,
  ...
}
```

The user, however, wouldn't see those names and, instead, I implement `Into<char>` for the respective characters you can imagine.

Presented with that problem, I could go for one of two solutions:
1. Use `EnumVariantNames`'s `<enum>::VARIANTS`, which is a static slice of strings. The problem with that approach is that it would generate an array with the names of the variants, which could be desired in many cases, but is not on mine as you can see above;
2. Use `EnumIter` and collect the result into a `Vec`. While the application would work fine with this approach, it seems overkill to heap-allocate a vector whose elements we know at compile time.

Considering those two points, I've decided to introduce this trait (and derive).

## Concerns

My only concern with this implementation is the name of both the trait/derive and the associated constant. I believe `VARIANTS` is a more fitting name, but it is already taken. I thought about renaming that to `VARIANT_NAMES`, but it feels like an unecessary breaking change, although I'd like to hear more opinions on this.

## Closing thoughts

This is my first public macro contribution and, while I think it's quite simple and small, I'd love to get feedback on the implementation!

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Strum has implemented the following macros:
 | [EnumMessage] | Add a verbose message to an enum variant. |
 | [EnumDiscriminants] | Generate a new type with only the discriminant names. |
 | [EnumCount] | Add a constant `usize` equal to the number of variants. |
+| [StaticVariantsArray] | Adds an associated `ALL_VARIANTS` constant which is an array of all enum discriminants |
 
 # Contributing
 
@@ -80,3 +81,4 @@ Strumming is also a very whimsical motion, much like writing Rust code.
 [EnumDiscriminants]: https://docs.rs/strum_macros/0.25/strum_macros/derive.EnumDiscriminants.html
 [EnumCount]: https://docs.rs/strum_macros/0.25/strum_macros/derive.EnumCount.html
 [FromRepr]: https://docs.rs/strum_macros/0.25/strum_macros/derive.FromRepr.html
+[StaticVariantsArray]: https://docs.rs/strum_macros/0.25/strum_macros/derive.StaticVariantsArray.html
--- a/strum/src/lib.rs
+++ b/strum/src/lib.rs
@@ -211,6 +211,16 @@ pub trait VariantNames {
     const VARIANTS: &'static [&'static str];
 }
 
+/// A trait for retrieving a static array containing all the variants in an Enum.
+/// This trait can be autoderived by `strum_macros`. For derived usage, all the
+/// variants in the enumerator need to be unit-types, which means you can't autoderive
+/// enums with inner data in one or more variants. Consider using it alongside
+/// [`EnumDiscriminants`] if you require inner data but still want to have an
+/// static array of variants.
+pub trait StaticVariantsArray: std::marker::Sized + 'static {
+    const ALL_VARIANTS: &'static [Self];
+}
+
 #[cfg(feature = "derive")]
 pub use strum_macros::*;
 
@@ -241,5 +251,6 @@ DocumentMacroRexports! {
     EnumVariantNames,
     FromRepr,
     IntoStaticStr,
-    ToString
+    ToString,
+    StaticVariantsArray
 }
--- a/strum_macros/Cargo.toml
+++ b/strum_macros/Cargo.toml
@@ -26,4 +26,4 @@ rustversion = "1.0"
 syn = { version = "2.0", features = ["parsing", "extra-traits"] }
 
 [dev-dependencies]
-strum = "0.25"
+strum = { path = "../strum" }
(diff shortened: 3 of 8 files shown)


Syndelis:
Apparently it fails CI because the new trait I added to `strum` is *obviously* not available on crates.io yet. Should I remove the doctest for the time being and introduce it in a new PR? How do you guys usually deal with this problem?

Syndelis:
Hi there, @Peternator7 ! Can I get some feedback on this one? I could split the tests into a separate PR, which would likely make the CI pass

image

Peternator7:
Hey @Syndelis , this looks really good! Don't worry too much about the CI if that's the only issue; it's been an annoying footgun with strum since early versions.

Peternator7:
What do you think about `VARIANTS` vs `ALL_VARIANTS`. It "conflicts" with the VariantNames derive, but that might be okay since type resolution should be able to figure out which one you want?

Syndelis:
> What do you think about `VARIANTS` vs `ALL_VARIANTS`. It "conflicts" with the VariantNames derive, but that might be okay since type resolution should be able to figure out which one you want?

I'm not a fan of this name either haha. I think best case scenario is renaming `VariantNames::VARIANTS` to `VariantNames::VARIANT_NAMES` and attributing `VARIANTS` to this new trait, but I understand this may not be worthy of a breaking change.

Other than that, I really don't know what name would best suit it. I've considered `STATIC_VARIANTS` and `VARIANTS_SLICE`, but it feels weird to add the type or storage into the constant's name. I'm open to suggestions!

polarathene:
This PR would remove the need for this snippet to have a const array/slice of all variants?:

```rs
// Strum presently lacks a derive for this:
const VARIANTS: &'static [MyEnum] = MyEnum::ALL_VARIANTS.as_slice();

// Effectively a const compatible equivalent of: MyEnum::iter().collect()
impl MyEnum {
    const ALL_VARIANTS: [Self; Self::COUNT] = {
        // Cannot rely on `Option<Self>` to default to `None`, as conversion
        // to output type is not possible? Thus use the first variant as a default:
        let default = Self::variant_from(0);
        let mut variants = [default; Self::COUNT];

        // Cannot use for loop (or iterators with collect) in a const fn,
        // Hence initializing an array and replacing default elements via while loop:
        let mut i = 0;
        while i < Self::COUNT {
            variants[i] = Self::variant_from(i);
            i+=1;
        }

        variants
    };

    // Must unwrap value due to no support for const traits,
    // otherwise could have collected with adaptors over [Option<Self>]:
    const fn variant_from(discriminant: usize) -> Self {
        let Some(variant) = Self::from_repr(discriminant) else { panic!("Invalid discriminant provided") };
        variant
    }
}
```

I was going to raise a feature request for that since it took a while to figure out a solution with all the `const fn` limitations 😅 

---

> Other than that, I really don't know what name would best suit it.

`VARIANTS_LIST`? Although I agree that the current `VARIANTS` would seem more appropriate. 

I think [this issue](https://github.com/Peternator7/strum/issues/198) is requesting the same feature here? It suggested `ITEMS`.

My snippet above could also be a `const fn` instead resulting in `MyEnum::variants()` / `MyEnum::list_variants()` / `MyEnum::get_variants()` / `MyEnum::itemize()`, etc. With that last one, it kinda relates to `ITEMS` suggestion 🤔

Syndelis:
Hi there, @Peternator7! Just wondering if you think anything else must be done to this PR or #315 in order to merge them?

nvzqz:
@polarathene Divan can now use runtime arguments via the [`args`](https://docs.rs/divan/latest/divan/attr.bench.html#args) option, so a custom `MyEnum` type would be usable now.

Peternator7:
Just published version [`0.26.0`](https://crates.io/crates/strum) on crates.io with this feature. I renamed the macro `VariantArray` and the associated constant `VARIANTS`. The type-inference system should be able to differentiate with VARIANTS you're referring to if they're both in scope. Thanks for your help on this feature!

Issue #299, opened 2023-09-28
EnumIter does not work for enums named 'Option'


horacimacias:
I'm using some code that has generated structs and enums, and some of the enumerations are called `Option`. They are on their own module/crate but there seems to be a conflict with the std::option::Option.

The following does work (note Options, not Option as the name of the enum)

```rust
#[derive(strum_macros::EnumIter, Debug)]
pub enum Options {
    Open,
    Closed,
}

use strum::IntoEnumIterator;
fn main() {
    for variant in crate::Options::iter() {
        println!("The door is {:?}", variant);
    }
}

```

the following does not work (the name of the enum is now Option)
```rust
#[derive(strum_macros::EnumIter, Debug)]
pub enum Option {
    Open,
    Closed,
}

use strum::IntoEnumIterator;
fn main() {
    for variant in crate::Option::iter() {
        println!("The door is {:?}", variant);
    }
}

```

PR #300, opened 2023-09-28
Fix EnumIter for enums named Option

horacimacias:
Fixes https://github.com/Peternator7/strum/issues/299


Peternator7:
LGTM, thanks for the fix!

PR #311, opened 2023-10-28
Impl Copy, Eq, and FusedIterator for EnumIter iterators

jwodder:
* The fields within `EnumIter` types (`usize` and `PhantomData`) all implement `Clone`, `Copy`, `Eq`, and `PartialEq`, so we can derive all of those on the iterator types.

* Once an `EnumIter` reaches the end of its values, it only ever yields `None`.  This means we can implement `FusedIterator` on it.

### DIFF
--- a/strum_macros/src/macros/enum_iter.rs
+++ b/strum_macros/src/macros/enum_iter.rs
@@ -71,9 +71,7 @@ pub fn enum_iter_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
 
     Ok(quote! {
         #[doc = #doc_comment]
-        #[allow(
-            missing_copy_implementations,
-        )]
+        #[derive(Clone, Copy, Eq, PartialEq)]
         #vis struct #iter_name #ty_generics {
             idx: usize,
             back_idx: usize,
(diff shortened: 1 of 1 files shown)


Peternator7:
Hey @jwodder, thanks for the PR! Is this blocking something specific that you're doing? I'm happy to add Clone, and FusedIterator; I'm skeptical of the rest however. Adding Copy will almost definitely cause someone to accidentally mutate a copy rather than the original, and semantically, PartialEq on an iterator seems like an unusual thing to do.

jwodder:
@Peternator7 

> Is this blocking something specific that you're doing?

No, it just seemed appropriate.

> Adding Copy will almost definitely cause someone to accidentally mutate a copy rather than the original

Fair.

> and semantically, PartialEq on an iterator seems like an unusual thing to do.

It doesn't seem unusual to me (though I happen to always derive `Eq` and `PartialEq` when possible).  On the other hand, I looked over some iterators in std, and the only ones I see that implement `Eq` are the range types from `std::ops`, which have non-iterator uses.

Feel free to edit the PR to remove the traits you don't want.

Peternator7:
Thanks for clarifying. After thinking about it, I've merged the FusedIterator addition, but decided to leave the other changes out. Thanks for the update!

PR #309, opened 2023-10-18
bump phf from 0.10 to 0.11

jqnatividad:

PR #319, opened 2023-12-22
chore(strum): update phf 0.10 -> 0.11

lovesegfault:


Peternator7:
Duplicate of #309, phf `0.11` bumped the msrv to `1.60`. I try to keep the msrv to the oldest release from the last 2 years which is currently `1.58`. Will merge 309 in april/may though when we get to 2 years since 1.60

PR #320, opened 2023-12-23
Add `attributes` to `strum_discriminants` for custom attributes in generated discriminant enum

0x5abe:
Sometimes derive macros need a specific attribute to be above them to work properly. I ran across this while using [bilge](https://crates.io/crates/bilge), where the [TryFromBits](https://docs.rs/bilge/0.2.0/bilge/derive.TryFromBits.html) derive macro requires the [bitsize](https://docs.rs/bilge/0.2.0/bilge/attr.bitsize.html) attribute above it. The `EnumDiscriminants` derive macro currently doesn't allow for arbitrary attribute macros before the generated enum's derives, so I added the `#[strum_discriminants(attributes(..))]` attribute to allow for this.

This PR is a draft because I am not entirely sure how a test for this specific attribute would be performed, and I wouldn't want it to get merged without its proper tests.

### DIFF
--- a/strum_macros/src/helpers/metadata.rs
+++ b/strum_macros/src/helpers/metadata.rs
@@ -3,8 +3,8 @@ use syn::{
     parse::{Parse, ParseStream},
     parse2, parse_str,
     punctuated::Punctuated,
-    Attribute, DeriveInput, Expr, ExprLit, Field, Ident, Lit, LitBool, LitStr, Meta, MetaNameValue,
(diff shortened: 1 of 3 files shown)


Peternator7:
Hey @SabeMP, I'd suggest the following. If we can find an example with `serde` where this would be helpful, I'd be happy to add that as a dependency to the tests crate so we can validate this works consistently.

Also, could you undo the formatting changes?

PR #321, opened 2023-12-25
fix: avoid conflicts in case of consecutive underscores

### LINKED ISSUE #225: Preserve leading underscores in generated `Ident`s
heck documents, under [Definition of a word boundary](https://docs.rs/heck/latest/heck/index.html#definition-of-a-word-boundary):

> Characters not within words (such as spaces, punctuations, and underscores) are not included in the output string except as they are a part of the case being converted to. Multiple adjacent word boundaries (such as a series of underscores) are folded into one. (“hello__world” in snake case is therefore “hello_world”, not the exact same string). Leading or trailing word boundary indicators are dropped, except insofar as CamelCase capitalizes the first word.

As a consequence, where heck is used to construct a Rust `Ident` that might only be valid due to its leading underscore (such as [here](https://github.com/Peternator7/strum/blob/master/strum_macros/src/macros/from_repr.rs#L96-L97)), the resulting `Ident` may be invalid.  See rust-lang/rust#98158 for a report of an arising ICE.

strum should ensure that, in such cases, leading underscores are preserved (or else document that such identifiers are unsupported).

aminya:
In `FromRepr` if two fields exist that their only difference is underscores, they are converted to the same discriminant name. This PR fixes the issue by not changing the case of the fields. 

Fixes #225

### DIFF
--- a/strum_macros/src/macros/from_repr.rs
+++ b/strum_macros/src/macros/from_repr.rs
@@ -83,7 +83,10 @@ pub fn from_repr_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
             },
         };
 
-        constant_defs.push(quote! {const #const_var_ident: #discriminant_type = #const_val_expr;});
+        constant_defs.push(quote! {
+            #[allow(non_upper_case_globals)]
+            const #const_var_ident: #discriminant_type = #const_val_expr;
+        });
         arms.push(quote! {v if v == #const_var_ident => ::core::option::Option::Some(#name::#ident #params)});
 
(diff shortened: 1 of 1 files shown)


Peternator7:
Hi @aminya, I'm worried this fix will trigger casing warnings with Clippy. I think this change is fine, but we need to `#[allow(non_upper_case_globals)]` or this will raise compiler warnings

aminya:
@Peternator7 Sure, fixed it in 
https://github.com/Peternator7/strum/pull/321/commits/c1e881526dbb4ab483c6786a61af1611fd57b831

PR #324, opened 2024-01-27
Update Changelog etc in prep for release

Peternator7:


strum_macros/Cargo.toml
```diff
@@ -26,4 +26,4 @@ rustversion = "1.0"
 syn = { version = "2.0", features = ["parsing", "extra-traits"] }
 
 [dev-dependencies]
-strum = { path = "../strum" }
+strum = { path = "../strum", version= "0.25" }
```
martin-g, line 29: Is this intentional or a typo ?
It might be intentional due to the cyclic dependency and the fact that `cargo publish` can't deal with it.
But it still uses 0.25 in `master` - https://github.com/Peternator7/strum/blob/master/strum_macros/Cargo.toml#L29

Peternator7: I haven't figured out how to deal with the cyclic dependency yet when publishing :/ It's just needed for the doc tests which is unfortunate. I had to publish `0.26.1` yesterday and I used the opportunity to get them in sync, but hadn't pushed those commits to master yet. This should be fixed/consistent now. Thanks for catching it.

PR #331, opened 2024-02-04
Add `#[strum(transparent)]` attribute

### LINKED ISSUE #258: Add `#[strum(transparent)]` argument
This crate is amazing for tackling a lot of boilerplate code and also ensuring variant name changes propagate to the trait implementations. I use `AsRefStr` and `EnumString` quite a lot.

However, a lot of times I've found myself in the situation of having something like this:
```rust
enum MyEnum {
    A,
    B,
    C,
    Other(String)
}
```

Afaik there's no way of propagating the `AsRef<str>` or `FromStr` implementations to the underlying `String`. What do you think about implementing a `transparent` argument to the attribute, similar to `#[serde(transparent)]` - only for one field provided in an unnamed/named fashion.

So then we could do:
```rust
#[derive(AsRefStr, EnumStr)]
enum MyEnum {
    A,
    B,
    C,
    #[strum(transparent)]
    Other(String)
}
```

I could try and pick this up if you like the idea. I think it would be handy for a lot of people.

bobozaur:
Adds the `transparent` attribute, usable with `AsRefStr`, `Display` and `IntoStaticStr`. This delegates the derived trait impl to the underlying (single) field of the enum variant.

Implementing this for `EnumString` would be nice, but it doesn't really act the same way as for the other traits. For `EnumString`, something like `#[strum(other)]` (similar to `#[serde(other)]`) would make more sense, because there you would have a last single attempt to parse the input into a `catch all` variant, whereas `trasparent` can be added to multiple variants.

Therefore, if desired, that should be handled in a separate PR.

Closes #258

### DIFF
--- a/strum/src/additional_attributes.rs
+++ b/strum/src/additional_attributes.rs
@@ -73,6 +73,10 @@
 //!     ```
 //!     The plugin will fail if the data doesn't implement From<&str>. You can only have one `default`
 //!     on your enum.
+//! 
+//! - `transparent`: Signals that the inner field's implementation should be used, instead of generating 
+//!    one for this variant. Only applicable to enum variants with a single field. Compatible with the 
+//!    `AsRefStr`, `Display` and `IntoStaticStr` derive macros.
 //!
 //! - `disabled`: removes variant from generated code.
 //!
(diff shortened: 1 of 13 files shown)


Peternator7:
This is great! Sorry for the long delay, but I've merged this and will hopefully get a new version out today

luxalpa:
> Therefore, if desired, that should be handled in a separate PR.

I would love to see this!

PR #334, opened 2024-02-17
feat(as_ref_str): enable `prefix` attribute for `AsRefStr` derive

vbrvk:
Enable `prefix` attribute for `AsRefStr` derive so

> (i.e. always `enum.as_ref().to_string() == enum.to_string()`).

Is true again


seanpianka:
Is this PR related to some bug in #296?

PR #345, opened 2024-03-11
Interpolate unnamed enum variant fields in to_string attribute

gin-ahirsch:
I'm thinking it would also make sense to emit an error, or at least a warning, if `{}`-brackets are found in the `to_string`-literal on unit variants. I'd be happy to add that.

### DIFF
--- a/strum_macros/src/macros/strings/display.rs
+++ b/strum_macros/src/macros/strings/display.rs
@@ -124,7 +124,7 @@ pub fn display_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
                         }
                     }
                 }
-                Fields::Unit => quote! { #name::#ident #params => ::core::fmt::Display::fmt(#output, f) }
+                Fields::Unit => quote! { #name::#ident #params => ::core::fmt::Display::fmt(&format!(#output), f) }
             };
 
             arms.push(arm);
--- a/strum_macros/src/macros/strings/display.rs
+++ b/strum_macros/src/macros/strings/display.rs
@@ -29,7 +29,20 @@ pub fn display_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
 
         let params = match variant.fields {
             Fields::Unit => quote! {},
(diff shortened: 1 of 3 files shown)


Peternator7:
Looks great! Only change is let's error on the empy `{}`. In the example from the PR: `"hue is {1}, saturation is {}"`, mixing them feels bad to me so let's just not allow it by disallowing the empty braces

gin-ahirsch:
What about disallowing `{}` on unit variants? Currently that will format "as-is", i.e. an `on_string="{}"` is like `format!("{{}}")` on a unit variant.

Peternator7:
Hey, sorry for the delay here. I don't feel too strongly either way. I'm inclined to reject it for consistency with other types of enums, but could be persuaded the other way.

Peternator7 (approved):

strum_macros/src/macros/strings/display.rs
```diff
@@ -29,7 +29,20 @@ pub fn display_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
...
-            Fields::Unnamed(..) => quote! { (..) },
+            Fields::Unnamed(ref unnamed_fields) => {
+                // Transform unnamed params '(String, u8)' to '(ref field0, ref field1)'
+                let names: Punctuated<_, Token!(,)> = unnamed_fields
+                    .unnamed
+                    .iter()
+                    .enumerate()
+                    .map(|(index, field)| {
+                        assert!(field.ident.is_none());
+                        let ident = syn::parse_str::<Ident>(format!("field{}", index).as_str()).unwrap();
```
tronical, line 40: I think this use of format!() broke the no_std builds. We see `error: cannot find macro `format` in this scope`. Should this perhaps be `alloc::fmt::format!`?

ogoffart: that format is fine as it is run at build time.
The problem is the format in the next chunk which is under a quote!

PR #351, opened 2024-04-18
fix cargo clippy warning and error

rtczza:
fix cargo clippy warning and error

### DIFF
--- a/strum_tests/tests/enum_is.rs
+++ b/strum_tests/tests/enum_is.rs
@@ -28,12 +28,12 @@ enum Foo {
 }
 #[test]
 fn generics_test() {
-    let foo = LifeTimeTest::One(Cow::Borrowed("Hello"));
-    assert!(foo.is_one());
-    let foo = LifeTimeTest::Two("Hello");
-    assert!(foo.is_two());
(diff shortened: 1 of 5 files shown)


strum_tests/tests/enum_is.rs
```diff
@@ -28,12 +28,12 @@ enum Foo {
 }
 #[test]
 fn generics_test() {
-    let foo = LifeTimeTest::One(Cow::Borrowed("Hello"));
```
Peternator7, line 31: Was clippy complaining about this? It doesn't on my machine, and I prefer we avoid changing variables names based on personal preference

Peternator7 (changes requested):
Thanks for fixing the clippy errors. I'd prefer we don't change the test case names unless they're getting flagged for something that I'm not seeing

strum_tests/tests/enum_is.rs
martin-g: I think it complains about `foo` only in `nightly`. I agree it could be ignored for now.

Peternator7:
Fixed in #473

PR #353, opened 2024-04-26
Add support for `const_into_str` attribute to enable static string conversions in const contexts

### LINKED ISSUE #352: Support for const_into_str to Enable Static String Conversion in Const Context
Currently, the `strum` library does not support converting enum variants into static strings within a const context. This limitation arises because the library uses the `From` trait, which cannot have const methods, even in nightly Rust. This issue becomes particularly significant when attempting to use [concatcp](https://docs.rs/const_format/latest/const_format/macro.concatcp.html) to concatenate two static strings in a const context.

To address this limitation, I propose the addition of a new enum meta attribute, `const_into_str`. This attribute would enable generating an additional public const function for enums, allowing conversion to static strings in a const context.

This enhancement would greatly improve `strum`'s functionality by allowing more flexible and performant use of enums in constant expressions. I have prepared an implementation and will submit a pull request for review.

biryukovmaxim:
This pull request introduces the `const_into_str` attribute to the `strum` macros, enabling enums to be converted into static strings in const contexts. This feature is particularly useful when working with compile-time string manipulations, such as those required by `concatcp`.

**Key Changes:**
1. Addition of `const_into_str` keyword in enum metadata processing.
2. Conditional generation of a `const fn into_str()` method for enums marked with `const_into_str`.
3. Adjustment in the generation of the `From` trait to utilize `into_str()` when `const_into_str` is enabled.

The implementation ensures that existing functionality is preserved and provides users with the option to opt into this new feature as needed. This feature could be especially beneficial for users needing efficient, compile-time safe string operations involving enums.

Please review the changes and let me know if there are any modifications or improvements required.

resolves #352

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,6 @@ Strumming is also a very whimsical motion, much like writing Rust code.
 [EnumDiscriminants]: https://docs.rs/strum_macros/latest/strum_macros/derive.EnumDiscriminants.html
 [EnumCount]: https://docs.rs/strum_macros/latest/strum_macros/derive.EnumCount.html
 [FromRepr]: https://docs.rs/strum_macros/latest/strum_macros/derive.FromRepr.html
-[VariantArray]: https://docs.rs/strum_macros/latest/strum_macros/derive.StaticVariantsArray.html
-[VariantNames]: https://docs.rs/strum_macros/latest/strum_macros/derive.EnumVariantNames.html
+[VariantArray]: https://docs.rs/strum_macros/latest/strum_macros/derive.VariantArray.html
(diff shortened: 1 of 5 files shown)


biryukovmaxim:
@Peternator7 could you check it out?

Peternator7:
Seems reasonable to me, could you add a test case?

biryukovmaxim:
> Seems reasonable to me, could you add a test case?

Added

lf94:
Could someone on the team merge this?

biryukovmaxim:
@Peternator7 could you take a look?

PR #360, opened 2024-06-02
Fix the unconditional usage of string interpolation in strum

### LINKED ISSUE #359: strum-macros 0.26.3 breaks `no_std` build
With 0.26.3, the emitted code when using `derive(strum::Display)` uses `format!` unconditionally. This requires the surrounding code to `use format::alloc;`, which seems like an unfortunate breaking change to me.

Example `main.rs`:

```rust
#![no_std]

#[derive(strum::Display)]
enum Foo {
    Bar
}

fn main() {
}
```

Sample `Cargo.toml`:

```toml
[package]
name = "strumtest"
version = "0.1.0"
edition = "2021"

[dependencies]
strum = { version = "0.26.2", features = ["derive"] }
strum_macros = { version = "0.26.3" }
```

If you downgrade to `=0.26.2` the build works, otherwise it aborts with:

```
error: cannot find macro `format` in this scope
 --> src/main.rs:3:10
  |
3 | #[derive(strum::Display)]
  |          ^^^^^^^^^^^^^^
  |
  = note: this error originates in the derive macro `strum::Display` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `strumtest` (bin "strumtest") due to 1 previous error
```

I think it's okay to require the surrounding `no_std` code to `use alloc::format;` when using `strum`, but it would be great if this requirement could come with a new major version (0.27) instead.

Peternator7:
fixes #359

### DIFF
--- a/strum_macros/src/macros/strings/display.rs
+++ b/strum_macros/src/macros/strings/display.rs
@@ -124,7 +124,17 @@ pub fn display_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
                         }
                     }
                 }
-                Fields::Unit => quote! { #name::#ident #params => ::core::fmt::Display::fmt(&format!(#output), f) }
+                Fields::Unit => {
+                    let used_vars = capture_format_strings(&output)?;
+                    if !used_vars.is_empty() {
+                        return Err(syn::Error::new_spanned(
+                            &output,
+                            "Unit variants do not support interpolation",
+                        ));
+                    }
+
+                    quote! { #name::#ident #params => ::core::fmt::Display::fmt(#output, f) }
+                }
             };
 
             arms.push(arm);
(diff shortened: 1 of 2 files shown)


strum_macros/src/macros/strings/display.rs
```diff
@@ -124,7 +124,17 @@ pub fn display_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
                         }
                     }
                 }
-                Fields::Unit => quote! { #name::#ident #params => ::core::fmt::Display::fmt(&format!(#output), f) }
+                Fields::Unit => {
```
gmryuuko, line 127: Related to #363 
Since `format!` is no longer used for `Unit`, please do not check the curly brackets on it.
This will cause the use of a single curly bracket to fail to compile, and two curly brackets cannot be escaped.
Or continue to use `format!` for `Unit`, and release it as a breaking change.

Issue #364, opened 2024-06-14
Fix itertools compatibility part 2. Release strum 0.26.3


joshka:
While strum_macros 0.26.3 fixes the incompatibility with itertools (https://github.com/Peternator7/strum/issues/358), projects that use the strum crate and the derive feature do not automatically pick this up when picking up dependabot updates. This means that manual steps (cargo update) need to be taken to update any packages that use strum 0.26.2 and itertools.

Put another way, strum 0.26.2 with the derive feature enabled is incompatible with itertools 0.13.0 because it requires any version greater than 0.26.2 of strum_macros. Instead strum is only compatible with itertools 0.13.0 when it uses 0.26.3 of strum_macros.

Suggested fix: release 0.26.3 pointing at strum_macros 0.26.3

PR #365, opened 2024-06-14
fix: require strum_macros 0.26.3

joshka:
This is required for compatibility with itertools 0.13.0 fixed in #358

Fixes: https://github.com/Peternator7/strum/issues/364


Peternator7:
Makes sense to me. Just pushed version `0.26.3` of strum to crates

joshka:
> Makes sense to me. Just pushed version `0.26.3` of strum to crates

Thanks - this compiles properly in a downstream repo :)

PR #368, opened 2024-07-18
Useful changes to `EnumString`

AlexSherbinin:
1. Added support for enums that consuming strings they were parsed from. 
2. Changed behaviour to derive `From` trait instead of `TryFrom` when default variant is defined(`TryFrom`'ll be defined automatically with `Infallible` as error type).

> [!NOTE]  
> Maybe a breaking change, because error type of `TryFrom` trait is changed from `ParseError` to `Infallible`.
> Enums that consuming strings was leading to compiler error, so there's no breaking change in that part

Changes usage example:
```rust
#[derive(Debug, Eq, PartialEq, EnumString)]
#[strum(serialize_all = "UPPERCASE")]
enum HttpMethod<'a> {
    Get,
    Post,
    #[strum(default)]
    Unrecognized(&'a str),
}
```

### DIFF
--- a/strum_macros/Cargo.toml
+++ b/strum_macros/Cargo.toml
@@ -23,7 +23,7 @@ heck = "0.5.0"
 proc-macro2 = "1.0"
 quote = "1.0"
 rustversion = "1.0"
-syn = { version = "2.0", features = ["parsing", "extra-traits"] }
+syn = { version = "2.0", features = ["parsing", "extra-traits", "visit"] }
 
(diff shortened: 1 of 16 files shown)


Peternator7 (changes requested):
@AlexSherbinin, thanks for the PR! Can you revert the formatting/style changes from this PR and include only the feature updates? They're making this PR hard to review, and the git blame becomes hard to read read if it's full of formatting changes.

PR #369, opened 2024-07-19
Fix disabled variants still appearing in variant names array

### LINKED ISSUE #244: Disabled variant still included in VariantNames
It is my understanding that the `#[strum(disabled)]` attribute should remove it from all the expansions. However, it seems that it is still included in the `VariantNames` trait implementation.

## Example code

```rust
#[derive(strum::IntoVariantNames)]
enum Fields {
    Field0,
    Field1,
    Field2,
    #[strum(disabled)]
    Unknown,
}
```

## Expected Output
```rust
impl ::strum::VariantNames for ResponseIdent {
        const VARIANTS: &'static [&'static str] = &[
            "Field0",
            "Field1",
            "Field2",
        ];
    }
```

## Actual Output
```rust
impl ::strum::VariantNames for ResponseIdent {
        const VARIANTS: &'static [&'static str] = &[
            "Field0",
            "Field1",
            "Field2",
            "Unknown
        ];
    }
```

Tar-Tarus:
This PR resolves #244 

It includes the following changes so far:
- Test to check for this issue when using the macro a65a225db6f43a69e798b046f2e0d33dea73b9da
- Fix to make sure the generated VARIANTS array only contains variants which are not disabled 8444b6d51a349cff6bc70037a771ba5336912d73

### DIFF
--- a/strum_macros/src/macros/enum_variant_names.rs
+++ b/strum_macros/src/macros/enum_variant_names.rs
@@ -20,10 +20,16 @@ pub fn enum_variant_names_inner(ast: &DeriveInput) -> syn::Result<TokenStream> {
 
     let names = variants
         .iter()
-        .map(|v| {
-            let props = v.get_variant_properties()?;
-            Ok(props
-                .get_preferred_name(type_properties.case_style, type_properties.prefix.as_ref()))
+        .filter_map(|v| {
+            let properties = v.get_variant_properties();
+            match properties {
+                Ok(props) if props.disabled.is_none() => Some(Ok(props.get_preferred_name(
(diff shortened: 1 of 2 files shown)


Peternator7:
Hi @Tar-Tarus, thanks for the PR. I agree this behavior is inconsistent, but I'm pretty hesitant to make any breaking changes around something pretty core like this without knowing what projects in might break. So let me think about it, but I think we might be locked into this behavior with the current attributes.

In the future, I'd like to make most of the attributes scoped to a specific macro so that this behavior is more flexible

Peternator7:
Closing this because it's too disruptive a breaking change

Issue #332, opened 2024-02-06
Customize error message for FromStr


WojciechMula:
It would be perfect if the error message can be customized.

Another thing I personally find user-friendly is enumerating possible values in error message; like: "wrong value of verb; possible values: GET, POST, PUT".

Peternator7:
This was fixed in #380. You can use `parse_error_ty` and `parse_error_fn` to use and create a custom error type :)

https://docs.rs/strum/latest/strum/derive.EnumString.html

Issue #221, opened 2022-04-16
Enhancement: Better ParseError


JRAndreassen:
Hi...

Love the crate...
I'd love it even more if the from_str() gave better error messages.

When I parse a file with serde I often get:
```
Failed: "Custom { field: \"Matching variant not found\" }"
```
It would be ooohhh so much better if it said:
```
Failed: Custom { field: \"Matching variant not found: [EnumName:'Value']\" }"
```

Perhaps change:
```rust
pub enum ParseError {
    VariantNotFound(String),
}

...
    fn from_str( s: &str,    ) ...
{
...
        _ => ::core::result::Result::Err(::strum::ParseError::VariantNotFound(format!("StructName[{}]", s))),
}
```

I know it's extra overhead, but ohhh so useful.
Since serde does not always give a good error, even with serde_path_to_error... 

Thanks
   JR

Peternator7:
Hello @JRAndreassen, glad to hear you like the crate! What does your `Deserialize` impl look like? I wasn't aware that serde would use a type's `FromStr` impl when deriving Deserialize

JRAndreassen:
Hi @Peternator7 ,

Thanks for getting back to me...

It does not use it directly...
But the handy crate "serde_with" makes it possible.
It eliminates the redundant serde attributes and code,  by telling serde to use "from_str"[DeserializeFromStr] and "Display"[SerializeDisplay] 

```toml
serde = { version = "^1.0", features = ["derive","rc"] }
serde_derive = "^1.0"
serde_json = "^1.0"
serde_path_to_error = "*"
serde_with = "*"

strum = "*"
strum_macros = "*"
```

```rust
#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, DisplayMore)]
#[derive(AsRefStr, IntoStaticStr, EnumString, EnumMessage, EnumIter, FromRepr)]
#[strum(ascii_case_insensitive)]
#[serde_as]
#[derive(SerializeDisplay, DeserializeFromStr)]
#[repr(u8)]
pub enum Units {
   #[strum(to_string = "BytesBandwidth", serialize = "Bandwidth", serialize = "OctetsBandwidth"")]
   BytesBandwidth = 0,
   #[strum(to_string = "BytesMemory", serialize = "Memory", serialize = "Mem")]
   BytesMemory = 1,
   /// [PRTG] bytes
   #[strum(to_string = "BytesDisk", serialize = "Bytes", serialize = "Disk")]
[...]

PR #380, opened 2024-09-18
feat: FromStr derive could support setting the error type

### LINKED ISSUE #91: FromStr derive could support setting the error type
When you derive on an enum, sometimes you want to implement some `From<XXXNotFound>` for some other error types, with maybe more explicit messages. It would be great if we could have an attribute that enables us to set the function that would be used to build the error from the original `&str`.

e.g.
```rust
#[derive(EnumString)]
#[strum(parse_err_ty = "SomeEnumNotFound", parse_err_fn = "some_enum_not_found_err")]
pub enum SomeEnum {
	A,
	B,
}

pub struct SomeEnumNotFound;
fn some_enum_not_found_err(_not_found: &str) -> SomeEnumNotFound { SomeEnumNotFound }
```

generating

```rust
impl FromStr for SomeEnum {
    type Err = SomeEnumNotFound;
    fn from_str(s: &str) -> Result<SomeEnum, Self::Err> {
        match s {
            "A" => A,
            "B" => B,
            other => some_enum_not_found_err(other),
        }
    }
}
```

If `parse_err_fn` is not specified but `parse_err_ty` is, we could fallback on `From<strum::ParseError>`.

JimChenWYU:
resolve #91 #332 #221

PR checker failed because #381


Peternator7:
This looks really good! Sorry for the delay reviewing it

strum_macros/src/helpers/mod.rs
```diff
@@ -13,6 +13,13 @@ use proc_macro2::Span;
 use quote::ToTokens;
 use syn::spanned::Spanned;
 
+pub fn missing_parse_err_attr_error() -> syn::Error {
+    syn::Error::new(
+        Span::call_site(),
+        "`parse_err_ty` and `parse_err_fn` attribute is both required.",
```
martin-g, line 19:
```suggestion
        "`parse_err_ty` and `parse_err_fn` attributes are both required.",
```

sebschrader:
Please reconsider using string literals for paths. Attributes nowadays support arbitrary expressions (including paths) instead of only string literals.

JimChenWYU:
> Please reconsider using string literals for paths. Attributes nowadays support arbitrary expressions (including paths) instead of only string literals.
> 
> One of the authors of `rust-analyzer` recently highlighted that using string literals for such things is very bad for IDE support, in his EuroRust [2024 presentation](https://eurorust.eu/2024/talks/powerful-macros-good-ide-support-no-headaches-choose-two/).

So you want like this?

```rust
#[derive(Debug, EnumString)]
#[strum(
    parse_err_fn = crate::custom::some_enum_not_found_err,
    parse_err_ty =crate::custom::CaseCustomParseErrorNotFoundError
)]
enum CaseCustomParseErrorEnum {
    #[strum(serialize = "red")]
    Red,
    #[strum(serialize = "blue")]
    Blue,
}

mod custom {
    #[derive(Debug, Clone, Eq, PartialEq, Hash)]
    struct CaseCustomParseErrorNotFoundError(String);
    impl std::fmt::Display for CaseCustomParseErrorNotFoundError {
        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
            writeln!(f, "not found `{}`", self.0)
        }
    }
    impl std::error::Error for CaseCustomParseErrorNotFoundError {}
    pub fn some_enum_not_found_err(s: &str) -> CaseCustomParseErrorNotFoundError {
        CaseCustomParseErrorNotFoundError(s.to_string())
    }
}
```

Peternator7:
@sebschrader, I think it's a reasonable point. I've opened #410 to change this behavior since it's not released yet.

PR #386, opened 2024-10-21
New Enum Derive Type (Enum Return Groups)

SCDevel:
Ok first off I made this for another project of mine that uses `strum` and wanted to play around with macros and avoid my other project. so I don't really plan on updating this anymore unless I run into issues with my other project. But anyways I figured you guys/girls might be interested in it so I wrote some documents and am creating this pull request.

I apologize for any "bad" code but I truly only code as a boredom/hobby thing.

Here is a TLDR of what this does. more info can be found in the `lib.rs -> EnumGroups()` as a doc string and playing in the tests.

TLDR:
Edit (fixed my terrible grammar):
This groups all of the variants by there arguments and generates a struct with variable for each grouped return type wrapped in a `Option<T>`. each variant auto implements a new  `.get_groups()` which returns the generated struct of types; which should populated with all `None` except for the one that applies to the variant its called from. that result should be `Some(T)`


Peternator7:
Hi @SCDevel, thank you for sending a PR. I think this is a neat plugin, but it seems somewhat specialized to whatever your other project is. With that in mind, I'm not interested in merging this PR, but appreciate you sharing what you've been working on.

PR #389, opened 2024-10-31
Inline generated functions

marc0246:
This results in a 3.08% performance increase in one of my real life workloads. I was using the `EnumIter` expansion, but there's no reason any of the other expansions shouldn't have the attribute (except formatting functions, as inlining those is generally bad practice, especially error displaying and debug-formatting). For iterators especially, and even more so in a hot hot loop like in my case, not inlining the iterator can be devastating. Note that not even fat LTO saved me in this case which I'm already using.


Peternator7:
Makes sense. Thanks for the PR 👍

Issue #313, opened 2023-11-11
`get_bool` and `get_int` in `EnumProperty` are undocumented and always return None


fritzrehde:
When looking at [`EnumProperty`'s documentation](https://docs.rs/strum/latest/strum/trait.EnumProperty.html) that `get_bool` and `get_int` both always returned `None`. 
My expectation of these way that I could set a bool in a property like so:
```rust
    #[strum(props(HideInHelp = "true"))]
    Variant,
```
and then I thought `"true"` would get parsed into a bool, and I could just check an `Option<bool>` with `get_bool` at runtime.
What I don't understand is why `get_bool` exists and always returns `None`. Is the feature not ready/implemented yet, and was accidentally added to the master branch? It should also probably be documented at least with a doc-string to prevent the confusion I ran into.
Thanks for the cool crate!

Milo123459:
bump. i have just tried implementing this to no avail, so was wondering if @Peternator7 had anything to say?

PR #391, opened 2024-11-20
Implement `get_int`/`get_bool` for properties

marxin:
Fixes: #313

I am also suggesting to use `i64` as return type for `get_int` in order to cover also negative numbers.

### DIFF
--- a/strum/src/lib.rs
+++ b/strum/src/lib.rs
@@ -182,7 +182,7 @@ pub trait EnumMessage {
 /// ```
 pub trait EnumProperty {
     fn get_str(&self, prop: &str) -> Option<&'static str>;
-    fn get_int(&self, _prop: &str) -> Option<usize> {
+    fn get_int(&self, _prop: &str) -> Option<i64> {
         Option::None
     }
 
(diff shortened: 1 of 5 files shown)


Peternator7:
Hey @marxin, thanks for the PR! This has been requested for quite a while. I tend to agree with @pali6 though that this would probably be better if accepted literals of the expected type rather than parsing the string.

In some ideal world, we could accept an expression `#[strum(props(my_prop = true || false))]` but that's difficult because we'd need to figure out the type. I think a better solution in the first version is to change the grammar so that `Prop` parses a [`Lit`](https://docs.rs/syn/latest/syn/enum.Lit.html) rather than a `LitStr` then we can do a pattern match in the macro to figure out what type of property we're looking at.

marxin:
Here we go. I've implemented the suggested approach for the two functions. Eventually, I decided to track the 3 categories using HashMap in `enum_properties_inner`, but using arrays (the original commit in this PR) might be possible. Which approach do you prefer?

Peternator7:
Looks great to me! thanks for taking the time to work on this :)

marxin:
Thanks for merging. May I ask you whether you plan to choke a new release any time soon?

Milo123459:
bump - @Peternator7 any plans for a new release soon?

Peternator7:
@Milo123459 You might've seen already but 0.27 has this change

PR #403, opened 2025-01-20
Use asserts instead of println in Display docstrings

sukhmel:
I've recently discovered that I was using strum incorrectly assuming that content of variants will be printed. In 0.26 the content **can** be printed but as this is not the default, I've decided that it may still be beneficial to clarify in the docs what to expect from the default option.

To better convey the results and make sure the user expectations are set right, I propose to use `assert_eq` + `format` instead of `println` as the latter is not very informative without specifying what should be printed as a result.


sukhmel:
I was hasty and missed an incorrect copy-paste. Now everything should pass

Peternator7:
Seems reasonable to me 👍

PR #407, opened 2025-01-27
Fix `Display` macro in `#![no_std]` programs

bugadani:
In `#[no_std]` `format!` does not exist, so the `to_string` helper was not usable.


Peternator7:
Nice, probably reduces an allocation in `std` as well.

PR #411, opened 2025-02-08
Per cargo docs, this should fix the circular dep when publishing strum

Peternator7:


martin-g:
@Peternator7 Could you please share the commands/script that you use to publish ?
Do you publish the whole workspace at once or one crate at a time ?
I maintain a similar project (one lib + one macros crates) and this circular dependency is really annoying.

Issue #307, opened 2023-10-13
`EnumString` should derive `From<&str>` instead of `TryFrom<&str>` when `#[strum(default)]` is present


MingweiSamuel:
If you derive `EnumString` on a enum with a `#[strum(default)]` it generates a `TryFrom<&str>` implementation which can never return an error.

PR #432, opened 2025-04-09
Support infallible EnumString for enums with a default variant

scovich:
Any enum with a `#[strum(default)]` variant has infallible parsing in practice, but the compiler does not know this because the error type is still `strum::ParseError`.

This PR proposes a non-breaking change to address the issue: Define a new type-level attribute `parse_infallible`, which instructs `EnumString` to use `std::convert::Infallible` for `FromStr::Err`, and to derive `From<&str>` (always) instead of `TryFrom<&str>` (1.34+). This allows e.g.
```rust
let variant = MyInfallibleEnum::from("some_unknown_variant")
```
and (in 1.82+)
```rust
let Ok(variant) = MyInfallibleEnum::from_str("some_unknown_variant");
```

Existing `EnumString` derivations remain unchanged unless the user adds the `parse_infallible` attribute. As a future breaking change, we could start inferring `parse_infallible` automatically even for enums that do not carry the annotation. If a breaking change is immediately acceptable, we don't need to define the new attribute at all, and this PR becomes simpler.

NOTE: `TryFrom` is still available (for 1.34+) with `parse_infallible`, because the `From<&str>` derived by `EnumString` activates a blanket implementation of reverse [`Into`](https://doc.rust-lang.org/std/convert/trait.Into.html#impl-Into%3CU%3E-for-T), which in turn activates a blanket implementation of [`TryFrom`](https://doc.rust-lang.org/std/convert/trait.TryFrom.html#impl-TryFrom%3CU%3E-for-T).

Also update docs and add new unit tests that exercise infallible parsing for both normal and PHF enums.

Fixes https://github.com/Peternator7/strum/issues/307

### DIFF
--- a/strum_macros/src/helpers/metadata.rs
+++ b/strum_macros/src/helpers/metadata.rs
@@ -19,6 +19,7 @@ pub mod kw {
     custom_keyword!(const_into_str);
     custom_keyword!(use_phf);
     custom_keyword!(prefix);
+    custom_keyword!(parse_infallible);
     custom_keyword!(parse_err_ty);
     custom_keyword!(parse_err_fn);
 
@@ -64,6 +65,7 @@ pub enum EnumMeta {
         kw: kw::parse_err_fn,
         path: Path,
     },
+    ParseInfallible(kw::parse_infallible),
     ConstIntoStr(kw::const_into_str),
 }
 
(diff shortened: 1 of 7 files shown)


Peternator7:
Sorry for the long delay on this. I've had a number of personal things in the last year and haven't been able to devote much time to this project so I appreciate the understanding.'

I decided to implement this with a small breaking change (automatically implement "From" if possible by default) rather than opting into it so I used a different PR, but I appreciate the design and this feature is now available in `0.28` if you need it :)

scovich:
Sounds great, thanks!

Issue #433, opened 2025-04-18
discriminants with empty enum


crop2000:
I generate code that sometimes results in empty enums. I still would like to be able to derive `EnumDiscriminants`. The problem is that this generated function cannot be compiled:
```
impl<'_enum> ::core::convert::From<&'_enum UIPassiveValue> for UIPassive {
    #[inline]
    fn from(val: &'_enum UIPassiveValue) -> UIPassive {
        match val {}
    }
}
```
I don't really know what the most consistent behaviour would be.
I made a draft in this commit 
https://github.com/Peternator7/strum/commit/efd6ff784cc1be606851b5f7f8ed57c87a9305c4
to make generate the following code:
```
impl<'_enum> ::core::convert::From<&'_enum UIPassiveValue> for UIPassive {
    #[inline]
    fn from(val: &'_enum UIPassiveValue) -> UIPassive {
        match val { _ => todo!(),}
    }
}
```

crop2000:
I guess instead of the match `panic!("Cannot be called because it is implemented on a empty enum")` is better.

PR #435, opened 2025-04-22
allow discriminants on empty enum

crop2000:
Solves issue #433


Peternator7:
Seems reasonable! Thanks for the fix!

PR #440, opened 2025-05-11
Implement a `suffix` attribute for serialization of enum variants 

### LINKED ISSUE #392: Add the `suffix` attribute
This is similar to #296 , but instead of prefixing str infront, we add it behind:

```rs
use strum_macros::AsRefStr;

#[derive(AsRefStr)]
#[strum(prefix = "path/to/file/", suffix = ".asset")]
enum AssetTypes {
    Filename,
}

fn main() {
    println!("{}", AssetTypes::Filename.as_ref());  // prints "path/to/file/Filename.asset"
}
```

This could be useful for using `enum` types for specifying file paths, among many other things.

amogh-dambal:
This PR implements a `suffix` attribute for serializing enums via `Display` and `AsRefStr`. Much of the logic was borrowed (lifted) from the implementation of the `prefix` attribute since the implementation mechanism is fundamentally similar.

This PR also adds corresponding tests and documentation of the new feature, in addition to further tests of the `prefix` attribute to clarify behavior.

fixes #392

### DIFF
--- a/strum_macros/src/helpers/metadata.rs
+++ b/strum_macros/src/helpers/metadata.rs
@@ -19,6 +19,7 @@ pub mod kw {
     custom_keyword!(const_into_str);
     custom_keyword!(use_phf);
     custom_keyword!(prefix);
+    custom_keyword!(suffix);
     custom_keyword!(parse_err_ty);
     custom_keyword!(parse_err_fn);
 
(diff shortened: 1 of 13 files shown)


Peternator7:
This looks really great! Thanks for the improvement :)

PR #443, opened 2025-05-25
Change enum table callbacks to FnMut.

ClaytonKnittel:


Peternator7:
Thanks for the PR! Just as an fyi, EnumTable is mostly unsupported so I would recommend using one of the other rust crates that does the same thing for any production scenarios.

PR #467, opened 2025-11-16
Fix clippy warnings

paolobarbolini:


Peternator7:
Fixes are included in #473. Thanks!

PR #476, opened 2026-02-22
Make TryFrom and FromStr infallible if there's a default

### LINKED ISSUE #255: `strum::EnumString` deriving `std::convert::TryFrom<&str>` means `From<&str>` cannot be implemented
In strum 0.23 the `EnumString` derive now implements `TryFrom<&str>` as well in #186

Problem with this is that `TryFrom<&str>` is automatically derived for `&str: Into<MyEnum>`: https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html#impl-TryFrom%3CU%3E-for-T

This means that after strum 0.22 it is no longer possible to implement `From<&str> for MyEnum` and/or `Into<MyEnum> for &str`. We would want to implement those especially if we have a `#[strum(default)]` field which would make parsing infallible (?).

Besides reverting, another fix would be to automatically implement `From`/`Into` if parsing is infallible

Same as #228 I imagine

Peternator7:
Fixes #255 and #307. the PR makes `EnumString` implement `From<&str>`  instead of `TryFrom` if there is a `#[strum(default)]` attribute on one of the variants.

Incorporates ideas from #432, but decided the breaking change is probably acceptable since it's more accurate to
say the parse is infallible when it is.

*If you come across this and want the old behavior, you can do this:

```rust
#[derive(EnumString)]
#[strum(parse_error_ty = strum::ParseError, parse_error_fn = make_error)]
pub enum Color {
    Red,
    Blue
}

fn make_error(x: &str) -> strum::ParseError {
    strum::ParseError
}
```

### DIFF
--- a/strum_macros/src/lib.rs
+++ b/strum_macros/src/lib.rs
@@ -34,15 +34,15 @@ fn debug_print_generated(ast: &DeriveInput, toks: &TokenStream) {
 
 /// Converts strings to enum variants based on their name.
 ///
-/// auto-derives `std::str::FromStr` on the enum (for Rust 1.34 and above, `std::convert::TryFrom<&str>`
-/// will be derived as well). Each variant of the enum will match on it's own name.
-/// This can be overridden using `serialize="DifferentName"` or `to_string="DifferentName"`
+/// auto-derives `std::str::FromStr` on the enum. Each variant of the enum will match on its own
+/// name. This can be overridden using `serialize="DifferentName"` or `to_string="DifferentName"`
 /// on the attribute as shown below.
-/// Multiple deserializations can be added to the same variant. If the variant contains additional data,
-/// they will be set to their default values upon deserialization.
+/// Multiple deserializations can be added to the same variant. If the variant contains additional
+/// data, they will be set to their default values upon deserialization.
 ///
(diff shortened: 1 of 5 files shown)


nazar-pc:
Very questionable decision IMHO. It is very confusing to a developer when invalid string results in successful conversion. If anyone wants that, they'd just do `.unwrap_or_default()` on the result instead in those rare special cases.

Most of the time though developers want to know whether the input was correct or not instead of silently getting the default value.

Peternator7:
Maybe the title of the PR isn't entirely clear. This PR clarified existing behavior more than changing anything in a meaningful way. By default, `EnumString` still generates `FromStr<T, ParseError>` so you can handle the error however you'd like. However, if you included `#[strum(default)]` on a variant, that would become the "default" output if none of the other options matched, and it would capture the input string. This was always infallible, but strum continued to generate `TryFrom` even though the conversion couldn't fail. This PR just changed the behavior in the infallible case to more accurately reflect that the conversion can't fail if you have applied `#[strum(default)]` to one of the variants. Otherwise, the behavior remains the same.

nazar-pc:
I think I didn't read it carefully enough and somehow assumed it was `#[default]` rather than `#[strum(default)]`. It is not as bad as I thought it was initially. Thanks for explaining.

PR #487, opened 2026-07-21
ci: Lock MSRV-compatible dependencies

cuviper:
Use the "fallback" resolver from newer Cargo to generate a `Cargo.lock`
with dependencies that are compatible with MSRV 1.71.


cuviper:
This should unblock #486, and in turn #485.

